PR #21, opened 2018-07-31
add toc and notational conventions; split out changes section

mikebrow:
Starting my review of the spec. 

commit 1: adds an OCI header, toc, and notational conventions. Also simplifies the introduction and scope sections.

commit 2: split out the changes section into a separate document.

commit 3: clean up heading levels..

commit 4: nits

Signed-off-by: Mike Brown <brownwm@us.ibm.com>


changes.md
```diff
@@ -0,0 +1,101 @@
+# OCI Distribution Specification Changes
+
+This doc should be updated when changes are made to the specification, indicating what is different.
```
vbatts: These are not intended to be _only_ breaking changes, but _all_ changes?

spec.md
```diff
@@ -3,193 +3,138 @@ title: "HTTP API V2"
...
-The new, self-contained image manifest simplifies image definition and improves security.
-This specification will build on that work, leveraging new properties of the manifest format to improve performance, reduce bandwidth usage and decrease the likelihood of backend corruption.
-
-For relevant details and history leading up to this specification, please see the following issues:
+# Open Container Initiative
+## Distribution Specification
+
+This specification defines an API protocol to facilitate distribution of images.
+
+The goal of this specification is to standardize container image distribution based on the specification for the [Docker Registry HTTP API V2 protocol](https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md).
```
vbatts: Would it not be useful to have pointers to the previous docker-registry v1 as historical defacto info?

spec.md
```diff
@@ -3,193 +3,138 @@ title: "HTTP API V2"
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119) (Bradner, S., "Key words for use in RFCs to Indicate Requirement Levels", BCP 14, RFC 2119, March 1997).
+
+The key words "unspecified", "undefined", and "implementation-defined" are to be interpreted as described in the [rationale for the C99 standard][c99-unspecified].
+
+An implementation is not compliant if it fails to satisfy one or more of the MUST, MUST NOT, REQUIRED, SHALL, or SHALL NOT requirements for the protocols it implements.
+An implementation is compliant if it satisfies all the MUST, MUST NOT, REQUIRED, SHALL, and SHALL NOT requirements for the protocols it implements.
+
+## Historical Context
+
+For relevant details and a history leading up to this specification, please see the following issues:
```
vbatts: oh, maybe historical defacto info could just be more added here.

changes.md
mikebrow: What do you think about saying "when non-trivial changes are made" or "With the exception of de-minimis modifications, this doc should be updated when changes are made.."

spec.md
dmcgowan: I don't think we should mention that in the intro. The protocol inherited very little from v1 registry. If we have a section about the background it could be included, as well as the layer design and move to content addressability.

spec.md
dmcgowan: `Background`? This needs to be filled out

spec.md
```diff
@@ -3,193 +3,138 @@ title: "HTTP API V2"
...
+This specification covers URL layout and protocols for interaction between a registry and registry client.
+Registry implementations MAY implement other API endpoints, but they are not covered by this specification.
 
-This includes the following features:
+This specification includes the following features:
 
 - Namespace-oriented URI Layout
 - PUSH/PULL registry server for V2 image manifest format
 - Resumable layer PUSH support
 - V2 Client library implementation
```
dmcgowan: This last item does not belong as part of `This specification`

changes.md
dmcgowan: The changes must include clarifications because any ambiguity may cause two implementations to differ in a way that the clarification breaks one and not the other. For example `Clarified expected behavior response to manifest HEAD request.` did cause breakages although one could have argued that the `HTTP` spec was already clear about the behavior of `HEAD` making the change trivial. We just need to be careful about what we define as "trivial" to account for this.

spec.md
```diff
@@ -209,9 +154,9 @@ The V2 registry API does not enforce this.
 The rules for a repository name are as follows:
 
 1. A repository name is broken up into _path components_.
-   A component of a repository name must be at least one lowercase, alpha-numeric characters, optionally separated by periods, dashes or underscores.
-   More strictly, it must match the regular expression `[a-z0-9]+(?:[._-][a-z0-9]+)*`.
-2. If a repository  name has two or more path components, they must be separated by a forward slash ("/").
+A component of a repository name must be at least one of lowercase, alpha-numeric characters, optionally separated by periods, dashes or underscores.
```
jonjohnsonjr: The original is unclear, but this change doesn't parse for me. What do you think about this?

`A component of a repository name must be comprised of lowercase, alpha-numeric characters (at least one), optionally separated by periods, dashes or underscores.`

spec.md
vbatts: I know we said no to the registry daemon. Though it was up in the air for enough of client side for basic validation.

spec.md
vbatts: I think it must begin with `[[:alnum:]]`, but can also include `[.-_]`

spec.md
vbatts: yeah fair.

changes.md
mikebrow: kk then de-minimis (such as spelling changes)

spec.md
mikebrow: Changing to: "- V2 Client (Consumer) requirements"

spec.md
mikebrow: Adding a TODO

spec.md
dmcgowan: Ok, let's not touch this and discuss in #23

spec.md
mikebrow: going with: "A component of a repository name must begin with one or more lowercase alpha-numeric characters. Subsequent lowercase alpha-numeric characters are optional and may be separated by periods, dashes or underscores."

vbatts:
Cool. Points addressed for now.
LGTM

PR #36, opened 2018-12-19
output documentation

vbatts:
This carries over the efforts from image-spec and runtime-spec to make
output documentation (PDF and HTML).

Also assumes a version of v0.1.0-dev for now.

### DIFF
--- a/.tool/genheader.go
+++ b/.tool/genheader.go
@@ -1,0 +1,54 @@
+// Copyright 2018 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"text/template"
+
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -1,5 +1,30 @@
 EPOCH_TEST_COMMIT	:= 91d6d8466e68f1efff7977b63ad6f48e72245e05
 
+DOCKER		?= $(shell command -v docker 2>/dev/null)
+PANDOC		?= $(shell command -v pandoc 2>/dev/null)
+
+OUTPUT_DIRNAME	?= output/
+DOC_FILENAME	?= oci-distribution-spec
+
+PANDOC_CONTAINER ?= docker.io/vbatts/pandoc:1.19.1-3.fc27.x86_64
```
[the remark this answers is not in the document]
vbatts: that would be nice. I have not investigated an org or team that we could share.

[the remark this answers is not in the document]
vbatts: I'm trying out https://quay.io/repository/oci/pandoc?tab=tags

jzelinskie:
LGTM

vbatts:
@opencontainers/distribution-spec-maintainers PTAL

.tool/genheader.go
```diff
@@ -0,0 +1,54 @@
+// Copyright 2017 The Linux Foundation
```
mikebrow: might want to change the year here..

version.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2016 The Linux Foundation
```
mikebrow: ditto

.tool/genheader.go
vbatts: hmmm, fair. It's fairly directly copied code, but I get it.

mikebrow: yeah one way to handle that is to put a year range or just add one for this year..

dmcgowan:
LGTM

PR #38, opened 2018-12-21
spec: upload UUID header

vbatts:
the `Docker-Upload-UUID` is currently optional, but vaguely described as
such. This makes it more clearly defined with a rename and sets the
`Docker-` header as legacy.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -429,7 +429,6 @@ If the POST request is successful, a `202 Accepted` response will be returned wi
 Location: /v2/<name>/blobs/uploads/<uuid>
 Range: bytes=0-<offset>
 Content-Length: 0
-Docker-Upload-UUID: <uuid>
 ```
 
 The rest of the upload process can be carried out with the returned url, called the "Upload URL" from the `Location` header.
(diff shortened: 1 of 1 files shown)


atlaskerr:
Is this change planned for V1?

dmcgowan:
LGTM

I think this is a safe change for v1 from the Docker/containerd perspective. This ID is used for resumable upload which is not yet implemented by the distribution client library or in the clients. Did you check if skopeo used this?

jzelinskie:
Comes up clean: https://github.com/containers/skopeo/search?q=Docker-Upload-UUID&type=Code

stevvooe:
Lgtm

I’m a little worried about rename for rename sake but this looks harmless.

jonjohnsonjr:
LGTM -- our clients rely on the `Location` header instead of `Docker-Upload-UUID`.

stevvooe:
After some thought, we might want to actually namespace this as `OCI-Blob-Upload-ID`.

stevvooe:
> LGTM -- our clients rely on the Location header instead of Docker-Upload-UUID.

This should be codified in the spec, if it not already is. It's a much better approach. We may want to drop the header, but I put it in there so that clients didn't have to parse the url when reporting error messages, which could cause reliance on the URL format.

Issue #22, opened 2018-08-29
[RFP] replace catalog API functionality


vbatts:
for more rich indexing and searching of container images in a registry. 
There is the `/v2/_catalog` though it still seems not clear enough for implementers.

jonjohnsonjr:
I'd like to remove `/v2/_catalog` from the spec entirely. It feels like an implementation detail that leaked into the docker spec, and it subverts the registry namespacing by being a global endpoint. My understanding is that it was an "admin API", which doesn't seem like it should be in scope of the distribution API, but a user-facing version of catalog would be great. 👍 

There are a couple relevant PRs in docker/distribution of the same vein that would be nice to include as well: 
* https://github.com/docker/distribution/pull/2169
* https://github.com/docker/distribution/pull/2199

From [the proposal](https://github.com/opencontainers/tob/blob/master/proposals/distribution.md), it's explicitly out of scope:

> Managing the grouping of image repository names is considered part of distribution policy or content management, which are out of scope. For example, “which image repositories are under library/?” is out of scope for this project.

... but it would be very nice to have, eventually :)

In general, I'd like to get the spec into a minimal, workable state before we start adding any features.

jonjohnsonjr:
With the caveat that **I think all of this is out of scope for this project**, I'd like to brain dump my thoughts on it so that we can maybe reach some consensus or have a plan for a proposal. Possibly, all of this could be a completely separate spec/service that many registry operators _just happen to_ host side-by-side with their distribution-spec compliant registry. That said...

Most<sup>[citation needed]</sup> registries don't implement `/v2/_catalog`, so it definitely shouldn't be a requirement. While I'd personally like to see this removed from the spec, there are some existing clients that consume this endpoint (I know of only [spinnaker](https://github.com/spinnaker/spinnaker), but there are likely others).

At the very least, we should mark this as [OPTIONAL](https://tools.ietf.org/html/rfc2119#section-5). (The rest of the spec would benefit from more formal Requirements Level language, too.)

Regardless of what we do here, it would be nice to have some method of indexing a registry that fits the spec's namespacing model. Being able to index the registry enables some nice projects, e.g. [flagstate](https://github.com/owtaylor/flagstate), [grafeas](https://github.com/grafeas/grafeas).

I haven't put together a formal proposal for anything yet, but some prior art to get the ball rolling:

### Listing Repostories

This + `/tags/list/` would enable clients to index at least the *tagged* images in a registry.

[...]

PR #40, opened 2019-01-04
spec: catalog operations optional.

atlaskerr:
This commit redefines the `_catalog` endpoint as an optional operation.

Background on the issue:
https://github.com/opencontainers/distribution-spec/issues/22
https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/rJ72OtZuhbc
https://github.com/opencontainers/tob/pull/35
https://github.com/opencontainers/tob/pull/46
https://github.com/opencontainers/tob/pull/50

Signed-off-by: Atlas Kerr <atlaskerr@gmail.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -718,7 +718,8 @@ Content-Type: application/json
 }
 ```
 
-Note that the contents of the response are specific to the registry implementation.
+Note that catalog operations are optional for registry implementations.
+Contents of the response are specific to the registry implementation.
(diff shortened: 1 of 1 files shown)


vbatts:
LGTM

dmcgowan:
LGTM

I think this is the right direction, but I would still like to consider having it completely removed from 1.0.  If I were to read this as the response is not standardized `Contents of the response are specific to the registry implementation`, then how can tools even leverage it. We could reserve parts of the URL namespace for further specification in the future, but for the most part registries should be free to add endpoints to add registry level operations.

PR #41, opened 2019-01-14
spec: fix binary JSON content-type

lucab:
This fixes endpoints examples to return the proper `application/json`
content-type for JSON content.

As per IETF specification and IANA registry below, the `application/json`
type is a binary media, so the content-type label does not need any
text-charset selector. Additionally, the media type definition
explicitly states that it has no required nor optional parameters,
which makes the current registry headers non-compliant.

Ref: https://www.iana.org/assignments/media-types/application/json

Signed-off-by: Luca Bruno <luca.bruno@coreos.com>

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -982,7 +982,7 @@ The registry does not implement the V2 API.
 401 Unauthorized
 WWW-Authenticate: <scheme> realm="<realm>", ..."
 Content-Length: <length>
-Content-Type: application/json; charset=utf-8
+Content-Type: application/json
 
 {
 	"errors:" [
@@ -1016,7 +1016,7 @@ The error codes that may be included in the response body are enumerated below:
 ```
 429 Too Many Requests
 Content-Length: <length>
-Content-Type: application/json; charset=utf-8
+Content-Type: application/json
 
 {
 	"errors:" [
@@ -1075,7 +1075,7 @@ The following parameters should be specified on the request:
 ```
 200 OK
 Content-Length: <length>
-Content-Type: application/json; charset=utf-8
+Content-Type: application/json
 
 {
     "name": <name>,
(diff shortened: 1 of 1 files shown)


lucab:
This originally comes from https://github.com/docker/distribution/pull/2813, and tries to solve the ambiguity of a non-standard content-type which was the root cause of this k8s+containerd+quay regression https://github.com/kubernetes/kubernetes/issues/72863#issuecomment-453926310.

@dmcgowan @vbatts PTAL.

vbatts:
LGTM

jzelinskie:
This should probably be cross posted, but here we go:

The details of the regression was that Quay added `; charset=utf8` to the Content-Type for _manifests_. This broke clients. However, for error messages, docker/distribution has always had the charset in addition to `application/json`. While this isn't technically correct, I'm not sure whether or not removing this field for v1 of OCI will break backcompat with certain clients. The tests for having that `charset` field are abundant in Docker distribution and I don't know of many of the clients are that diligent about checking for it.

stevvooe:
LGTM

This was pulled from examples on other apis. Clients that choke on this have a bug anyways, so I’m not sure claiming this is the root cause is quite right. Correct handling on the client will parse out the field and ignore it if it doesn’t apply.

jzelinskie:
I think most of the code that checks for the charset was written by you, @stevvooe. So if you think it's fine, I'll rubber stamp it.

My experience with working with http headers in Go is that nothing in net/http helps you with the values pulled out of headers (e.g. splitting on `;`). Thus, I suspect that docker clients might choke, but this is something we should fix sooner rather than later.

lucab:
Bump, does this need more green-stamps or can it be merged already?

jzelinskie:
LGTM

PR #44, opened 2019-01-16
OpenAPI Documentation

atlaskerr:
This PR adds OpenAPI document for the distribution spec along with the accompanying jsonnet used to generate it. Build commands have also been added to the Makefile.

Eventually the spec can be used to generate markdown documentation with the help of golang's `text/template`.


atlaskerr:
To view the spec, go to [http://petstore.swagger.io](petstore.swagger.io) and paste in the following link:

https://raw.githubusercontent.com/atlaskerr/distribution-spec/openapi/openapi.json

ad-m:
@atlaskerr: What do you think about adding multiple tags for different resources (```catalog```, ```blob```, ```manifest``` etc.) for readability purpose & grouping ( see https://swagger.io/docs/specification/grouping-operations-with-tags/ )?

openapi/operations.libsonnet
```diff
@@ -0,0 +1,280 @@
...
+    .addResponse(resp.err.forbidden)
+    .addResponse(resp.err.notFound)
+    .addResponse(resp.err.tooManyRequests)
+  ,
+
+  local blobExists =
+    op.new(
+      operationId='oci-blob-exists',
+      tags=tags,
+      summary='Check for the existence of a manifest by name and reference.',
```
ad-m: ```Check for the existence of a blob by name and reference.```

openapi/paths.libsonnet
```diff
@@ -0,0 +1,110 @@
...
+
+local items = {
+  oci:: {
+    local base =
+      pathItem.new()
+      .addOperation('GET', ops.oci.verify),
+
+    local catalog =
+      pathItem.new()
+      .addParameter(params.n)
```
ad-m: Why string? Why required?

openapi/vendor/oci-schemas/content-descriptor.schema.json
```diff
@@ -0,0 +1,63 @@
...
+         "type": "string"
+      },
+      "mediaType": {
+         "description": "",
+         "pattern": "^[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}/[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}$",
+         "type": "string"
+      },
+      "size": {
+         "description": "This REQUIRED property specifies the size, in bytes, of the raw content. This property exists so that a client will have an expected size for the content before processing. If the length of the retrieved content does not match the specified length, the content SHOULD NOT be trusted.",
+         "maximum": 9223372036854775808,
```
ad-m: Why?

openapi/vendor/oci-schemas/content-descriptor.schema.json
```diff
@@ -0,0 +1,63 @@
...
+         "minimum": -9223372036854775808,
```
ad-m: Why? There is any other specs which reference to that limit?

openapi/vendor/oci-schemas/errors.schema.json
```diff
@@ -0,0 +1,26 @@
+{
+   "$id": "http://opencontainers.org/schema/error",
+   "$schema": "http://json-schema.org/draft-07/schema#",
+   "description": "This property contains a list of errors messages returned from the server due to an invalid request.",
+   "properties": {
+      "errors": {
+         "items": {
+            "properties": {
+               "code": {
+                  "type": "string"
```
ad-m: enum?

openapi/vendor/oci-schemas/fs.go
```diff
@@ -0,0 +1,2611 @@
...
+`,
+	},
+
+	"/": {
+		name:  "/",
+		local: `.`,
+		isDir: true,
+	},
+
+	"/.git": {
```
ad-m: We really need that?

openapi/vendor/oci-schemas/content-descriptor.schema.json
ad-m: Och, I don't have a lot experience and I don't notice that it's vendored library.

openapi/vendor/oci-schemas/errors.schema.json
atlaskerr: The distribution spec doesn't supply an official jsonschema for errors. Let me double-check but I believe the schema is correct.

atlaskerr:
@ad-m  Thanks for the advice! I updated the spec with the appropriate tags.

openapi/vendor/oci-schemas/fs.go
atlaskerr: This comes from another package but thanks for pointing that out. I'll clean up this up in the other repo.

openapi/paths.libsonnet
atlaskerr: There isn't an official jsonschema for catalog. It's an optional endpoint so I'm not too worried about it.

openapi/vendor/oci-schemas/errors.schema.json
ad-m: I agree that it is correct, but it can be even more precise, because the official documentation contains a enumeration of possible error codes.

openapi/operations.libsonnet
atlaskerr: Fixed.

openapi/vendor/oci-schemas/content-descriptor.schema.json
ad-m: Should a negative value be allowed?

openapi/vendor/oci-schemas/errors.schema.json
atlaskerr: Ah you're right. Totally forgot about that. I'll work on it!

openapi/paths.libsonnet
ad-m: The specification must be accurate and precise. I believe that this parameter should be optional. Is there any docs mention ```n``` as required? 

The JSONSchema specification can be used for generated API clients, and then the optional parameter makes huge difference for compilator, code, library usage etc.

ad-m: Documentation mention:
```
The catalog for a given registry can be retrieved with the following request:

GET /v2/_catalog
```

It mean there is no required parameter like ```n``` or ```limits```.

atlaskerr: Sounds good. I'll work on this one as well.

openapi/vendor/oci-schemas/content-descriptor.schema.json
atlaskerr: Probably not. Let me double-check the image-spec's schema and get back to you.

atlaskerr:
@ad-m Is this your first time looking at jsonnet code? Were you able to read the code OK?

ad-m:
@atlaskerr No, I do not have experience with JSONNet. In recent weeks, I've worked a lot with OpenAPI to open the HyperOne.com Platform API (api.hyperone.com/docs is our docs). I got to know the optimal form of specifications to build the constraints of generators, because - in our case - OpenAPI specification is not intended for man, so it must be especially precise. We are interested in this specification, as we are hope basing on this generate image registry flexible enough to our usage.

openapi/vendor/oci-schemas/content-descriptor.schema.json
atlaskerr: https://github.com/opencontainers/image-spec/blob/befc3cba3ffdf76e2f72f52a248a20a2727cc4c9/schema/defs.json#L19

Image-spec allows for negative numbers. I agree that it should not have negative numbers though. I'll hold off on fixing this until I can make a PR in the image-spec.

openapi/paths.libsonnet
atlaskerr: Fixed.

openapi/vendor/oci-schemas/errors.schema.json
atlaskerr: Enums added.

ad-m:
In order to verify whether the specification is useful and fulfills its task, it recommends attempts to generate the client in several languages. Then you will see in a simple way what information in the specification is missing & discovery how to reshape specification to make it more useful eg. remove all inline objects. See for example at https://github.com/OpenAPITools/openapi-generator.

atlaskerr:
@ad-m I've fixed the deprecated field so it doesn't show up.

> Please, don't repeat yourself. Response Object are Map[string, Header Object | Reference Object], so please use Reference Object instead

I'm not repeating myself. The main reason why I used jsonnet was  so I wouldn't have to use reference objects.  I create an object once in jsonnet and reference it throughout the spec. It's much easier to read this way.

ad-m:
Who is it easier to read such OpenAPI specs (7,3k+ lines so far)? What is the output that we want to provide? OpenAPI specifications in a popular format or OpenAPI generator in JSONNet format?

Do you think that the OpenAPI specification is not sufficient and needs to be improved? Have you proposed these changes to the OpenAPI specification to include JSONNet in OpenAPI v4 specification directly? What's wrong with JSONSchema references?

What is the purpose of a OpenAPI specification that does not contain a data model at all (```.components.schemas``` are empty), but only a data structure?

atlaskerr:
@ad-m Openapi docs aren't mean't to be read by humans. That's why the spec is defined in json. Users can use their favorite openapi viewer to read the spec. I'm not going against openapi. The jsonnet generates a valid openapi spec.

ad-m:
@atlaskerr Does OpenAPI mean only visualizations of the API structure for people for you? Should the developer in case of communication problems not refer to the specifications in the original format?

How do you expect a programmer to track changes in the specification (maintance specification in time)?

How should they analyze situations such as discrepancies between specifications and the answers given? The JSONSchema ecosystem provides automatic tools to those that work more efficiently when the data is modeled correctly.

Have you experience in generating client libraries that allow you to use such an API? This greatly improves interoperability, because it allows you to get API libraries for many of languages. 

See for example OpenAPI schema ( https://api.hyperone.com/openapi.json ) which generates visualization for humans ( https://api.hyperone.com/docs ), but however, it is also used to generate an automatically updated, always up-to-date API library ( https://github.com/hyperonecom/h1-client-go , https://github.com/hyperonecom/h1-client-js ). Process of generation API libraries is work-in-progress.  

Docker Engine has the correct OpenAPI specifications ( https://docs.docker.com/engine/api/v1.30/ ). It allows to generate / create always up-to-date clients eg. https://www.npmjs.com/package/docker-engine . It require eg. keep data model in OpenAPI specification too.

atlaskerr:
@ad-m You're more than welcome to make a PR to change implementation details for the generated code after this PR is merged. This PR is to introduce a valid openapi document that fully describes the distribution spec. Would you say that's fair?

ad-m:
@atlaskerr I am interested in developing effective API documentation, which is why I spend time to accurately describes what changes in our specification are desirable to use full potential of OpenAPI ecosystem.

atlaskerr:
Alright the openapi doc is good to go and ready to be merged!

atlaskerr:
I'm closing this PR.

I gauged interest for the addition of an OpenAPI document for the distribution spec over a month and a half ago on the mailing list but got little feedback. I went ahead an made a PR anyway with the hopes that seeing the layout in the repo would spark interest but it doesn't seem like it's gonna pan out. Thanks to all that reviewed.

mikebrow:
Most of the folks were busy with end of year projects and conferences.. then there was the holiday break. If you like we can discuss the benefits of OpenAPI on one of the community calls (pros/cons, dependencies, demo, maintenance requirements, any alignment issues between cncf/oci and OpenAPI at the project level?, ...). Was there an issue opened to discuss this?

Cheers, Mike

vbatts:
atlas, I know this was a load of work. I have wrangled with jsonSchema and openAPI before. Always with big hopes, but consistently found that it became overkill, and unfortunately unless it became a strict gating process that was central to CI and validation then it soon became stale and ignored. 
I'm closing this PR as there is not consensus to make this approach a central validation. I do greatly appreciate your effort in investigating this.

atlaskerr:
@vbatts No worries. I feel the same way. I had fun working on it! :)

PR #45, opened 2019-01-17
spec: Remove catalog from spec.

atlaskerr:
This PR removes all references to the catalog endpoint from the distribution spec.

Maintainer approval
- [x] @dmcgowan
- [x] @jzelinskie
- [x] @mikebrow
- [ ] @stevvooe
- [x] @vbatts

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -30,7 +30,6 @@ The goal of this specification is to standardize container image distribution ba
   - [Content Digests](#content-digests)
   - [Pulling An Image](#pulling-an-image)
   - [Pushing An Image](#pushing-an-image)
-  - [Listing Repositories](#listing-repositories)
   - [Listing Image Tags](#listing-image-tags)
   - [Deleting an Image](#deleting-an-image)
 - [Detail](#detail)
@@ -56,8 +55,6 @@ The goal of this specification is to standardize container image distribution ba
     - [PATCH Blob Upload](#patch-blob-upload)
     - [PUT Blob Upload](#put-blob-upload)
     - [DELETE Blob Upload](#delete-blob-upload)
-  - [Catalog](#catalog)
-    - [GET Catalog](#get-catalog)
 
 ## Notational Conventions
 
@@ -700,113 +697,6 @@ The response format is as follows:
         ]
     }
 
-### Listing Repositories
-
-Images are stored in collections, known as a _repository_, which is keyed by a `name`, as seen throughout the API specification.
-A registry instance MAY contain several repositories.
-The list of available repositories is made available through the _catalog_.
-
-The catalog for a given registry can be retrieved with the following request:
-
-```HTTP
-GET /v2/_catalog
-```
-
-The response will be in the following format:
-
-```HTTP
-200 OK
-Content-Type: application/json
-
-{
-  "repositories": [
-    <name>,
-    ...
-  ]
-}
-```
-
-Note that catalog operations are OPTIONAL for registry implementations.
-Contents of the response are specific to the registry implementation.
-Some registries MAY opt to provide a full catalog output, limit it based on the user's access level or omit upstream results, if providing mirroring functionality.
-Subsequently, the presence of a repository in the catalog listing only means that the registry MAY provide access to the repository at the time of the request.
-Conversely, a missing entry SHOULD NOT mean that the registry does not have the repository.
-More succinctly, the presence of a repository only guarantees that it is there but not that it is _not_ there.
-
-For registries with a large number of repositories, this response MAY be quite large.
-If such a response is expected, one SHOULD use pagination.
-A registry MAY also limit the amount of responses returned even if pagination was not explicitly requested.
-In this case the `Link` header will be returned along with the results, and subsequent results can be obtained by following the link as if pagination had been initially requested.
-
-For details of the `Link` header, please see the [_Pagination_](#pagination) section.
-
-#### Pagination
-
-Paginated catalog results can be retrieved by adding an `n` parameter to the request URL, declaring that the response SHOULD be limited to `n` results.
-Starting a paginated flow begins as follows:
-
-```HTTP
-GET /v2/_catalog?n=<integer>
-```
-
-The above specifies that a catalog response SHOULD be returned, from the start of
-the result set, ordered lexically, limiting the number of results to `n`.
-The response to such a request would look as follows:
-
-```HTTP
-200 OK
-Content-Type: application/json
-Link: <<url>?n=<n from the request>&last=<last repository in response>>; rel="next"
-
-{
-  "repositories": [
-    <name>,
-    ...
-  ]
-}
-```
-
-The above includes the _first_ `n` entries from the result set.
-To get the _next_ `n` entries, one can create a URL where the argument `last` has the value from `repositories[len(repositories)-1]`.
-If there are indeed more results, the URL for the next block is encoded in an [RFC5988](https://tools.ietf.org/html/rfc5988) `Link` header, as a "next" relation.
-The presence of the `Link` header communicates to the client that the entire result set has not been returned and another request MUST be issued.
-If the header is not present, the client can assume that all results have been received.
-
-> __NOTE:__ In the request template above, note that the brackets are REQUIRED.
-> For example, if the url is `http://example.com/v2/_catalog?n=20&last=b`, the value of the header would be `<http://example.com/v2/_catalog?n=20&last=b>; rel="next"`.
-> Please see [RFC5988](https://tools.ietf.org/html/rfc5988) for details.
-
-Compliant client implementations SHOULD always use the `Link` header value when proceeding through results linearly.
-The client MAY construct URLs to skip forward in the catalog.
-
-To get the next result set, a client would issue the request as follows, using the URL encoded in the described `Link` header:
-
-```HTTP
-GET /v2/_catalog?n=<n from the request>&last=<last repository value from previous response>
-```
-
-The above process SHOULD then be repeated until the `Link` header is no longer set.
-
-The catalog result set is represented abstractly as a lexically sorted list, where the position in that list can be specified by the query term `last`.
-The entries in the response start _after_ the term specified by `last`, up to `n` entries.
-
-The behavior of `last` is quite simple when demonstrated with an example.
-Let us say the registry has the following repositories:
-
-    a
-    b
-    c
-    d
-
-If the value of `n` is 2, _a_ and _b_ will be returned on the first response.
-The `Link` header returned on the response will have `n` set to 2 and last set to _b_:
-
-```HTTP
-Link: <<url>?n=2&last=b>; rel="next"
-```
-
-The client can then issue the request with the above value from the `Link` header, receiving the values _c_ and _d_.
-Note that `n` MAY change on the second to last response or be fully omitted, depending on the server implementation.
 
 ### Listing Image Tags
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -692,116 +690,6 @@ The response format is as follows:
...
-More succinctly, the presence of a repository only guarantees that it is there but not that it is _not_ there.
-
-For registries with a large number of repositories, this response may be quite large.
-If such a response is expected, one should use pagination.
-A registry may also limit the amount of responses returned even if pagination was not explicitly requested.
-In this case the `Link` header will be returned along with the results, and subsequent results can be obtained by following the link as if pagination had been initially requested.
-
-For details of the `Link` header, please see the [_Pagination_](#pagination) section.
-
-#### Pagination
```
mikebrow: there's more detail in the catalog pagination section than in the pagination for tag results.. instead the tag pagination refers to this section.. so maybe filter through and see what can move down to the tags pagination section...

atlaskerr:
Alright, I'll take a look at it tomorrow.

spec.md
dmcgowan: It looks like at least `last` and `Link` referred to this section from there

jzelinskie:
Did we vote on this yesterday? I think Stephen wanted to keep it in the spec as optional, but I honestly think that just not having it in OCI 1.0 is equivalent since most registries will still support it for backwards compatibility with v2.2.

mikebrow:
catalog should probably remain as optional, till it's been replaced

vbatts:
On 31/01/19 11:32 -0800, Jimmy Zelinskie wrote:
>Did we vote on this yesterday? I think Stephen wanted to keep it in the spec as optional, but I honestly think that just not having it in OCI 1.0 is equivalent since most registries will still support it for backwards compatibility with v2.2.

I don't see a point in keeping it at all

mikebrow:
From: https://github.com/opencontainers/distribution-spec/issues/22#issuecomment-417147051

> @dmcgowan - I am +1 to not including the _catalog API at all. If we are going to have an OPTIONAL api, maybe we can have something that is more manageable, such as _list at any level, including /_list (lists first level namespaces if supported), /<repo>/tags/_list, <repo>/manifests/_list, and /<firstnamespacepart>/_list (lists repositories if supported). This is similar to the proposal you linked and we have more flexibility if we are not tied to _catalog.

I like what Derek proposed. I think being able to iterate over the contents is a fundamental operation. Simply removing catalog... without replacement, essentially removes our ability to iterate.

jonjohnsonjr:
I think if we don't want people to rely on catalog, we should remove it from the spec. Removing it from the spec won't make registries that currently support it suddenly stop, and I certainly don't want anyone else to implement it. It seems that we all agree it's bad and needs to be replaced by something else.

Both `_catalog` and `/tags/list` aren't really sufficient to build real systems on top of, as outlined here: https://github.com/opencontainers/distribution-spec/issues/22#issuecomment-417143466

I _would_ like to see some progress on a proposal for replacements to these things, but IMO both are out of scope for the **distribution** spec. If we want extensions for discovery and content management, we can do that separately.

Catalog (or something like it) is explicitly out of scope, per [the proposal](https://github.com/opencontainers/tob/blob/master/proposals/distribution.md):
> Managing the grouping of image repository names is considered part of distribution policy or content management, which are out of scope. For example, “which image repositories are under library/?” is out of scope for this project.

atlaskerr:
A solid compromise would be to keep `/v2/_catalog` optional and mark it as `DEPRECATED` with some kind of warning that talks about its flaws.

atlaskerr:
In order to make catalog usable in the future, I think are gonna have to have a long talk about how to logically group a collection of repositories into a spec-defined namespace. This will for sure be a breaking change so I don't forsee us being able to have a functional catalog endpoint until V2.

My 2 cents.

mikebrow:
Listing (and/or management) of repository names "is explicitly out of scope."  

I want a recount on that original decision... throws up hands... :-)

mikebrow:
@atlaskerr if you can just refactor the pagination text.. I think you'll have support for merging the PR. Cheers, Mike

atlaskerr:
My rebase messed up my branch and Github automatically closed it. Oops. I'll bring it back up when I fix the issue!

atlaskerr:
This PR is ready for review.

spec.md
```diff
@@ -700,94 +697,85 @@ The response format is as follows:
         ]
     }
 
-### Listing Repositories
 
-Images are stored in collections, known as a _repository_, which is keyed by a `name`, as seen throughout the API specification.
-A registry instance MAY contain several repositories.
-The list of available repositories is made available through the _catalog_.
+### Listing Image Tags
```
mikebrow: ?? I think you wanted to pick up the interesting description of how pagination works and merge that down into the pagination description in image tags.  Not move some portions of the description of image tags into the description of listing repositories. See references to repositories in the pagination examples below.

spec.md
```diff
@@ -700,94 +697,85 @@ The response format is as follows:
...
 
-Paginated catalog results can be retrieved by adding an `n` parameter to the request URL, declaring that the response SHOULD be limited to `n` results.
-Starting a paginated flow begins as follows:
+Paginated tag results can be retrieved by adding an `n` parameter to the request URL, declaring that the response SHOULD be limited to `n` results.
+
+Starting a paginated flow MAY begin as follows:
 
 ```HTTP
-GET /v2/_catalog?n=<integer>
+GET /v2/<name>/tags/list?n=<integer>
```
vbatts: wait, is this new?

vbatts: oh no, just consolidating examples ..

dmcgowan:
+1 on removal. Do we need a "reserved" endpoint function though to ensure this endpoint does not get redefined as something else in the future? I don't foresee it being reused but we should prevent collision through redefinition and ensure that repositories starting with `_` is forbidden.

stevvooe:
We do need to reserve this. If a registry wants to implement _catalog, then go forth.

BTW, we also need to deprecate `/list` for `/`, in certain APIs.

vbatts:
ok y'all, I've been hearing churning from several sources that while the `_catalog` route was a pain, that it is a depended on feature. At this point it does not matter that "it should have never been used", but it does exist and will be a blocker for any migration from what was the Docker Registry API v2.s2 over to what will be the OCI distribution spec v1.
If we do add routes to facilitate listing (and I do like some of the proposals that have been surfaced), it sounds like they will have to be be additive and not replacing of the `_catalog` route.
I'm a NACK for removing this.

SteveLasker:
It would be good to understand who’s dependent upon this and how it would carry forward. Although ACR implements it, I’m still an advocate for replacing. Removing it from the spec doesn’t preclude some distribution implementations from keeping their apis. ACR would likely keep it until we complete the new API. We’d then start a deprecation process. 
But, keeping it suggests others must implement it, and it doesn’t appear to be realistic so we’d have a spec that isn’t supportable. 
Can we queue this up for discussion next week.

cyphar:
To be clear, the plan for transition from Docker distribution to OCI distribution should be the same as the plan for transition from Docker images to OCI images -- the absolute minimum number of changes should be made that would require clients to be rewritten.

This will allow us to have an OCI project (even though it might be imperfect) that users can easily switch to. After that, we can go through normal spec change iteration and have a `1.1` (or `2.0`) which removes cruft we don't need and so on. Otherwise we will (further) needlessly delay the widespread adoption of OCI.

Removal of `_catalog` and replacing it with something else falls into a change that "would require clients to be rewritten".

jzelinskie:
>while the `_catalog` route was a pain, that it is a depended on feature

Only for some definition of "depended on". Programs like Spinnaker and Twistlock use `_catalog` to discover all of the content in a registry. However, registry implementations don't expose all images via the endpoint making that assumption false.

The problem of integration with tools that want to index a registry efficiently needs to be solved at post-Distribution v1.

>At this point it does not matter that "it should have never been used", but it does exist and will be a blocker for any migration from what was the Docker Registry API v2.s2 over to what will be the OCI distribution spec v1.

My argument is to because it is broken and people are mistakenly using it, the spec should have it as either optional or drop it under the assumption that every popular registry also implements Docker v2-2 in the meantime while a real solution can be created. Regardless of the spec, the popular registries are going to be stuck supporting this broken endpoint for a while.

SteveLasker:
i agree we want to make it easy for people to move to OCI. From what i understand, _catalog isn’t implemented by many existing distributions. There’s no client API in docker.exe or other common tooling that i'm aware of. 
I’m not suggesting a distribution that implements this or any other additive api wouldn’t be OCI compliant, but requiring the _catalog api in the spec says it is required, yet won’t be implemented. 
So, i’m just trying to understand why we would keep a spec on something that isn’t supported, yet can continue to be supported by those that have. 
Removing it frees up the space for a new api. Which, we do need more diligence in completing. (myself included in this self recognition of lack in progress)

SteveLasker:
On the scanner comment from Jimmy; this is true. And while we want to standardize for scanners and other more generic tooling, each scanner company has already had to deal with each clouds unique implementation, including different apis altogether. Removing the catalog API just frees up the mental model to construct something that each of us can implement consistently, reducing the churn on the scanners and other tooling.

amouat:
According to tweets I've seen from people at DockerCon, there is now a `docker registry ls` command. I can only assume this uses the catalog API (please correct me if anyone knows better!). This means that if we remove the catalog stuff, OCI registries will need to support a separate API to be fully compatible with the docker CLI.

There's still a strong argument to remove it, but I just wanted to flag this in case people weren't aware.

jonjohnsonjr:
It looks like `docker registry ls` is scoped to a repository, so it wouldn't need catalog, just tag listing.

SteveLasker:
it’s also a plug-in as part of Docker Enterprise 3.

amouat:
> it’s also a plug-in as part of Docker Enterprise 3.

What do you mean? It's not in the normal Docker CLI? Or it doesn't work against the regular docker distribution registry?

SteveLasker:
Apparently it’s only in Docker EE and only works with DTR, because the catalog and listing apis are so inconsistently implemented. ;). I’ve asked the docker folks to join us in helping define a new api we can all support these common scenarios: https://hackmd.io/s/BJPAUxDvV#OCI-Catalog-Listing-API---Workgroup

dmcgowan:
This seems stalled and perhaps we should put it to a vote of whether to include the catalog in 1.0 or not. I believe the maintainers are in agreement that we will not be adding any new endpoints for 1.0 for this functionality but wait for 1.1. I am still +1 for dropping in 1.0.

The `docker registry` plugin is something experimental to interact with DTR and it is not relevant to this conversation. However it would be good to have them (I am not involved in that) included in any discussion for creating a post 1.0 endpoint which could generically work for everyone.

SteveLasker:
While at KubeCon, several of us got together to discuss in person:
[Notes:](https://hackmd.io/s/HJdQ5HnTV)
## Summary 
As docker recently announced a new search api, implemented in DTR and Docker EE, because the `_catalog` api wasn't consistently supported, we agreed to remove `_catalog` from the spec. We agreed to reserve the endpoint name, to avoid conflict for registries that did implement `_catalog`

dmcgowan:
I put up a vote count in the comment. Let's get full approval then we can ask that this be rebased.

dmcgowan:
@mikebrow sounds good, if everyone agrees we can carry this PR or make sure that gets included before merging

vbatts:
we came to a conclusion on the call today that removes and reserves catalog API, as even currently all the registries that implement a catalog endpoint all do it differently. A complaint registry does not need to implement this endpoint, but registries that have already implemented such can keep it. There is no expected behavior for this endpoint

PR #48, opened 2019-01-31
Add Go structs

atlaskerr:
This PR adds Go structs for schemas defined by the distribution spec. There's not many but it will be easier to reason about adding more custom schemas in the future.


vbatts:
@dmcgowan this almost would be nice to carry over the most basic types from the distribution repo. Thoughts?

vbatts:
LGTM

PR #49, opened 2019-01-31
Add jsonschemas

atlaskerr:
This commit adds jsonschemas defined in the distribution spec.

Test fixture generation, file embeding, and schema validation have been
added to the Makefile as well.

Signed-off-by: Atlas Kerr <atlaskerr@gmail.com>


vbatts:
Big effort here on getting this prepped. The biggest thing I worry about here, because it bit us in the image-spec, is ensuring that every tweak in the spec language was lock-step reflected in the jsonschema, and that unit/integration testing ensuring there were no regressions along the way.

The other (more unfortunate) factor that played in was: many of the data structures are directly related to Golang structures and marshalling.

Every language handles marshelled structs differently and inevitably there is something that is cleaner in one language and inversely more difficult to handle in another. 

Thoughts there?

schema/errors.libsonnet
```diff
@@ -0,0 +1,76 @@
...
+  errors: 'This property contains a list of errors messages returned from the server due to an invalid request.',
+};
+
+// Error Schema
+local errors(output=JSV7) = {
+
+  local code = {
+    type: 'string',
+    enum: [
+      'BLOB_UNKNOWN',
```
stevvooe: This is why I don't like this stuff. What if we add a new error code? This validator fails if it encounters that. This is not an enum necessarily. There are set of known values and unknown future values.

schema/errors.libsonnet
```diff
@@ -0,0 +1,76 @@
+//   Copyright 2018 The Linux Foundation
```
stevvooe: Shouldn't this be the OCI Authors?

schema/errors.libsonnet
```diff
@@ -0,0 +1,76 @@
...
+      'NAME_UNKNOWN',
+      'SIZE_INVALID',
+      'TAG_INVALID',
+      'UNAUTHORIZED',
+      'DENIED',
+      'UNSUPPORTED',
+    ],
+  },
+  local message = { type: 'string' },
+  local detail = { type: 'string' },
```
stevvooe: Detail can be anything, not just a string.

Makefile
```diff
@@ -15,3 +15,20 @@ install.tools: .install.gitvalidation
...
+schemas:
+	find schema -path 'schema/test-fixtures' -prune -o -name generate.jsonnet -execdir jsonnet -J jsonnet -m . {} \;
+
+.PHONY: embed-schemas
+embed-schemas:
+	printf "%s\n\n%s\n" "$$(cat .header)" "$$(go generate schema/gen.go)" > schema/fs.go
+
+.PHONY: test-fixtures
+test-fixtures:
+	find schema -path 'schema/test-fixtures/**' -name generate.jsonnet -execdir jsonnet -J jsonnet -m . {} \;
```
stevvooe: Why use find here?

Makefile
```diff
@@ -15,3 +15,20 @@ install.tools: .install.gitvalidation
 
 .install.gitvalidation:
 	go get -u github.com/vbatts/git-validation
+
+# Generate json-schemas.
+.PHONY: schemas
+schemas:
+	find schema -path 'schema/test-fixtures' -prune -o -name generate.jsonnet -execdir jsonnet -J jsonnet -m . {} \;
```
stevvooe: Again, just use make input variable for the target. No need to use find.

stevvooe:
Why do we need jsonschema to validate like three fields? This PR adds 1000 lines of code to define the errors and catalog endpoint, and I can't really tell if its right or not. It also strictly defines the error codes and details type, which aren't an exact set.

Makefile
atlaskerr: This was a copypasta from my other project where I had a bunch of `generate.jsonnet` files in the schema directory. I can fix that.

atlaskerr:
@vbatts Yeah keeping everything in lockstep is kind of a drag but I think it's important to have json stuff defined using some kind of standard. Do you think it would be better to have an `oci-schemas` repository that mantains jsonschemas for all OCI specs?

> Why do we need jsonschema to validate like three fields? This PR adds 1000 lines of code to define the errors and catalog endpoint, and I can't really tell if its right or not. It also strictly defines the error codes and details type, which aren't an exact set.
I hear you. It is kind of overkill. I made this PR back when I believed distribution needed to define more custom schemas instead of just exposing the image-spec stuff so I went ahead and setup the foundation.

schema/errors.libsonnet
jzelinskie: There are multiple ways to handle this in JSONSchema using `anyOf` or `oneOf`.

jzelinskie:
Are we getting a lot of value out of the layer of indirection using jsonnet?
I'm not opposed to having a JSONSchema validation, but I think we need to KISS.

atlaskerr:
> Are we getting a lot of value out of the layer of indirection using jsonnet?

I used jsonnet simply because it's my default tool for working with json. I like having JSONSchema validation but would prefer not to have to work with raw json moving forward.

vbatts:
tbh jsonschema is painful either way. The biggest thing is to not increase the new formats to learn and maintain

vbatts:
atlas, I know this was a load of work. I have wrangled with jsonSchema and openAPI before. Always with big hopes, but consistently found that it became overkill, and unfortunately unless it became a strict gating process that was central to CI and validation then it soon became stale and ignored. 
I'm closing this PR as there is not consensus to make this approach a central validation. I do greatly appreciate your effort in investigating this.

PR #51, opened 2019-01-31
*: Add license header template.

atlaskerr:


mikebrow:
.header? the other .go files have this header in it already. Is this used by some tooling?

atlaskerr:
The .header is used to add license headers to generated code. The jsonschema PR converts oci schemas into a Go HTTP filesystem. I agree that this seems useless by itself lol. I'll close this and add it to the jsonschema PR so it's easier to reason about.

mikebrow:
> The .header is used to add license headers to generated code. The jsonschema PR converts oci schemas into a Go HTTP filesystem. I agree that this seems useless by itself lol. I'll close this and add it to the jsonschema PR so it's easier to reason about.

figured it was something like that.. maybe throw that stuff in a hack/ directory or whatnot and substitute the year for current year when it's generated / used...

PR #54, opened 2019-02-02
concepts: Documentation for Pulling Images

atlaskerr:
This PR is a proof-of-concept for separating the distribution spec's conceptual documentation from the endpoint reference.

The new `concepts` directory contains high-level workflows for working with the distribution spec. It is intended for users that want to know how a registry works but don't necessarily want  to implement one. The current spec is nearly 5K lines long and is super technical.

If this PR gets merged, I'll add more workflows.


stevvooe:
@atlaskerr What is wrong with the introduction section of the specification? How about we revise that and break out the API reference?

atlaskerr:
I'll make another PR for revising the introduction.

PR #61, opened 2019-04-22
Correct grammar for digest algorithm portion

fabianfreyer:
The current grammar specifies that the characters `[A-Fa-f0-9_+.-]` are valid in the `algorithm` portion of a digest. This contradicts the specification stating that:
> compliant implementations SHOULD use sha256

`sha256` does not match the above rule.

This pull request fixes this issue by additionally allowing characters `g-zG-Z` in the `algorithm` portion.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -234,7 +234,7 @@ We define a _digest_ string to match the following grammar:
 
 ```EBNF
 digest      := algorithm ":" hex
-algorithm   := /[A-Fa-f0-9_+.-]+/
+algorithm   := /[A-Za-z0-9_+.-]+/
 hex         := /[A-Fa-f0-9]+/
 ```
 


jzelinskie:
The intention is to use hexadecimal for the checksum part of the string. The regular expression is for handling hexadecimal and not SHA256 specifically. In what scenario would the characters after `f` or `F` be used?

fabianfreyer:
The problem is that the current grammar states the algorithm part also has to be hex.

dmcgowan:
@fabianfreyer good catch, you are right

@jzelinskie this only changes the algorithm, the content still remains hex. I agree with you about keeping that the same.

jzelinskie:
LGTM

fabianfreyer:
Thanks!

Issue #26, opened 2018-09-14
vendor-specific strings in HTTP headers


dongsupark:
The current distribution-spec contains vendor-specific strings like `Docker` in HTTP headers. Though it's not clear to me what we should do about it, because simple string substitutions would be incompatible changes. So I'm simply listing them below, without creating a PR.

* `Docker-Content-Digest`
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/changes.md
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#digest-header
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#existing-manifests
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#existing-layers
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#completed-upload
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#cross-repository-blob-mount
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#on-success-ok-3
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#on-success-created
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#on-success-ok-4
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#on-success-temporary-redirect
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#on-success-accepted-1
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#on-success-upload-complete

* `Docker-Distribution-API-Version`
  * https://github.com/opencontainers/distribution-spec/blob/59d293a11d862b2aad0c984dc6516f0401830045/spec.md#api-version-check

* `Docker-Upload-UUID`
[...]

PR #64, opened 2019-06-18
spec: replace "Docker-" mnemonics

rchincha:
Considering that this spec is dealing with OCI layouts, is there any reason to leave "Docker-" mnemonics in the API contracts?


hallyn:
Looks good.

jonjohnsonjr:
There was some discussion here: https://github.com/opencontainers/distribution-spec/issues/26

tl;dr, it's probably not worth the headache of breaking existing clients

dmcgowan:
This specification aims to standardize and be compatible with as many existing registries as possible. For 1.0 we are rejecting any changes which reduce that compatibility. We are accepting new OPTIONAL changes which do not break existing clients/registries.

Issue #27, opened 2018-10-10
The specification needs some formalization


mikebrow:
certain language like "must" should be converted to "MUST" when describing an absolute requirement of the specification.

atlaskerr:
I'm gonna tackle this one this week, hopefully by Thursday.

jzelinskie:
Fixed in #46.

mikebrow:
Additional formalization is required above just upper casing the prior must type verbage.

mikebrow:
Short list skimming over the first 250 lines.

Example 1 (line 192): ```All client implementations SHOULD treat unknown error codes as UNKNOWN, allowing future error codes to be added without breaking API compatibility. For the purposes of the specification error codes will only be added and never removed.```

Should->MUST

Example 2 (line 206): ```The client SHOULD be prepared to ignore this data.```

Informal, and incomplete.. should be rewritten.

208-209 needs to be reconciled with the earlier statements about authentication.. 

211 ```If `404 Not Found` response status, or other unexpected status, is returned, the client SHOULD proceed with the assumption that the registry does not implement V2 of the API.```

1) or other unexpected status? is there a list ..  2) remove should and assumption.. use MUST language.

213-215 ```When a `200 OK` or `401 Unauthorized` response is returned, the "Docker-Distribution-API-Version" header SHOULD be set to "registry/2.0". Clients MAY require this header value to determine if the endpoint serves this API. When this header is omitted, clients MAY fallback to an older API version.```

1) not clear why we would talk about older api versions .. 2) needs to be resolved with the docker/oci table when done. 3) SHOULD and MAY language should be revised here  think.

246 ```compliant implementations SHOULD use sha256```

SHOULD -> MUST at least support sha256.. ???

mikebrow:
need to address: `<!--- TODO: add relevant background information here --->`

mikebrow:
skimming over the error section.. there's a lot of MAY and SHOULD language that seems to render the processing of said error "indefinite"  

Consider rewriting to make these more definite, with MUST language and explain more specifically what is OPTIONAL.

atlaskerr:
@mikebrow  Ugh it seems like I'm gonna have to use my brain on this one. I'll open a PR today and ping you  later this week when a roughdraft is ready.

mikebrow:
There's a lot of "will be" language throughout to be revised.  The Issue I think is that the spec was written from a what the client should expect. Versus what the Registry and Client MUST implement.

[...]

PR #66, opened 2019-07-26
Add registry mirroring section

### LINKED ISSUE #12: spec: support for passing client image name for mirroring use case
There is an assumption to today that a server implementation of the distribution specification will either not care about the name used by the client or that all requests will have a known common namespace. An example of this is the Docker Hub assuming that all requests are prefixed with `docker.io` even though the registry hostname is `registry-1.docker.io`. However this has always caused difficulty when a client then wants to mirror content, `localhost:5000/library/ubuntu` could proxy to `registry-1.docker.io/library/ubuntu`, however `localhost:5000` could never proxy to anything else. Complicated registry configurations have been proposed to remedy this as well as a backwards incompatible approach of requesting as `localhost:5000/docker.io/library/ubuntu`.  However a goal of this specification should be simplicity and backwards compatibility. I believe that a solution does belong in the specification to unlock the mirroring use cases without complicated configuration or DNS setup.

My proposal is to add a way to pass up the name resolved by the client to the registry, (e.g. `docker.io/library/ubuntu`). So if a request is going to `localhost:5000/library/ubuntu`, it could mirror both `docker.io/library/ubuntu` and `quay.io/library/ubuntu` and switch based on request parameters. There are 2 possible ways to achieve this, one is by creating adding an HTTP request header (e.g. `OCI-REF-NAME: docker.io/library/ubuntu`) or by adding a query parameter `?oci-ref-name=docker.io/library/ubuntu`). The first is clean but the second may be more useful for static mirroring. I am not suggesting one over the other yet, just stating the problem and solutions to discuss.

dmcgowan:
Define repository namespace query parameter for mirroring.

Closes #12


spec.md
```diff
@@ -901,6 +902,20 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
...
 
+### Registry Mirroring
+
+A registry MAY operate as a mirror for an upstream registry to support pull-through caching or proxying of fetch operations (such as fetching tags, manifest, or blobs).
+In this case, the `Host` HTTP Header passed to the mirror likely does not match the repository's namespace used by the client.
+The `ns` query parameter on fetch operations is OPTIONAL and specifies the repository's namespace used by a client.
+The repository's namespace is normally the first component of a full repository name used by a client.
+This namespace is often used to determine the registry host a client will interact with, however, a client MAY be configured to use a different host.
+A mirrored registry MAY use the `ns` query parameter to resolve an upstream registry host.
+A mirrored registry MAY choose to use a single upstream registry and ignore the `ns` query parameter.
```
mikebrow: /s/A mirrored/The mirror/
/s/use a single upstream registry and//

spec.md
```diff
@@ -901,6 +902,20 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
...
+
+A client SHOULD be aware of whether a registry host being used is a mirror.
```
mikebrow: /s/host/host is/

spec.md
```diff
@@ -901,6 +902,20 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
...
+A client SHOULD avoid sending `ns` query parameters to non-mirrored registries.
```
mikebrow: /s/non-mirrored/non-mirror/

spec.md
```diff
@@ -901,6 +902,20 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
...
+A client SHOULD NOT send authorization credentials for an upstream registry to a mirrored registry.
```
mikebrow: /s/mirrored/mirror/

spec.md
```diff
@@ -901,6 +902,20 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
 
 > for more details, see: [compatibility.md](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#backward-compatibility)
 
+### Registry Mirroring
+
+A registry MAY operate as a mirror for an upstream registry to support pull-through caching or proxying of fetch operations (such as fetching tags, manifest, or blobs).
```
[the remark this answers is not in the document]
dmcgowan: Fine, I'll update to pull. That is more consistent

jzelinskie:
Why do clients need to know anything about pull-through caching if its implemented server-side?

dmcgowan:
> Why do clients need to know anything about pull-through caching if its implemented server-side?

The clients should know how the registry host was resolved from a given image reference. The clients don't care how the server is implemented, but they SHOULD provide information to the server which indicates what the reference being asked for is. Just as when an HTTP client connects with a PROXY server it must communicate what the upstream server is, the same is true here. Today the protocol doesn't define anyway to communicate what the upstream is and proxies end up be hardcoded to a single upstream. In a few cases you can see proxies use custom domains per upstream and require users to change the name of their images in order to use them.

josephschorr:
> Today the protocol doesn't define anyway to communicate what the upstream is and proxies end up be hardcoded to a single upstream.

Right... isn't that the point? If I encode that "myregistry/mynamespace/myrepo" goes to "upstream/foo/bar", that's a detail for the maintainer of `myrepo` and one that the client, ideally, doesn't need to know; the whole point is the puller thinks they are getting "myregistry/mynamespace/myrepo".

If the goal is to allow the *client* to specify "upstream/foo/bar", then I'd that the target is not really a repository anymore, but simply a working proxy, and thus, a different protocol parameter might be useful, but registries should therefore have the option to not support said parameter.

dmcgowan:
> Right... isn't that the point? 

That is one use case that will still work. In the example you mentioned, when a repository is proxied in that fashion, the puller often does know of this detail as they must explicitly provide myregistry with the intent of getting some upstream content. The use case where myregistry is some sort of blessed version of upstream is reasonable, but not the intent of the namespace parameter here.

> If the goal is to allow the client to specify "upstream/foo/bar"

This is the use case here and proxy may be better terminology here, but that is really a detail of the registry. The registry may act as a proxy, proxy-cache, or active mirror, that is out of scope for definition here. This parameter just enables all of those features to work across multiple namespaces. For example if you want public images from both `docker.io/*` and `quay.io/*` to be cached in the same registry proxy today, you would need the server to have two hostnames (something like docker.io.myproxy and quay.io.myproxy) then have clients configured to do that mapping for each namespace. This simple query parameter provides a much simpler option to clients and servers. If a server does not support it, it ignores the parameter. If a client is configured to send all requests to a server which does not support it, that is not different than any other misconfiguration by clients today.

josephschorr:
> For example if you want public images from both docker.io/* and quay.io/* to be cached in the same registry proxy today, you would need the server to have two hostnames (something like docker.io.myproxy and quay.io.myproxy) then have clients configured to do that mapping for each namespace

Or configure two repositories, one for each? (especially since combining them could lead to merge conflicts).

I'm concerned we're adding quite a bit of complexity to address a use case that has simpler solutions when configured on the registry side.

dmcgowan:
> I'm concerned we're adding quite a bit of complexity to address a use case that has simpler solutions when configured on the registry side.

Can you elaborate here? Configuring a repository for each mirror is non-trivial. Configuring a domain for each upstream and routing to the upstream based on the domain is not easier, that would still requires the same routing on the server side that an implementation of this would require. The client side implementation to support per-registry configuration is not simple and inherently requires catch -all conditions when trying to enforce proxying through a gateway.

I did do a [client side implementation](https://github.com/containerd/containerd/pull/3451) of this to demonstrate the feature and allow server side implementations a client to test against. On the client side, it is not complex at all since clients should already know how to handle 404s when multiple registry endpoints are configured. On the server side, the complexity to support this isn't much more than existing proxy-cache support.

josephschorr:
> Configuring a repository for each mirror is non-trivial.

Its non-trivial, but its not **that** difficult either :)

My concern remains around complexity: the document as outlined, for example, says that the `ns` should not be sent to non-mirroring registries... but how does the client know that? Is it the registry's job to report back an error if that argument is found but unsupported? How will clients know to be able to check for this capability?

If we feel that pass-through proxying of other registries is, in and of itself, a feature of the **protocol** (rather than something configured on the registry side), then I suspect we need to give significantly more thought to the end-to-end user experience. For example, I could imagine some paths supporting proxying and others not.

dmcgowan:
> how does the client know that?

The clients have the most context and really does not need to be defined here, only that a client SHOULD make that distinction to avoid sending unnecessary redundant information. The clients themselves have both the configuration and endpoint resolution logic, so it has multiple options for determining this. In the implementation I sent I just simply did this by checking whether the endpoint was configured without push support, as this could indicate the registry being communicated with may not be the upstream source. However, I will probably add a check there for `ns != host` since there are never push configurations (such as with a Kubernetes runtime). Either way, this is trivial and not required.

> Is it the registry's job to report back an error if that argument is found but unsupported?

No, the registry can simply ignore it. This is like asking a registry today which was configured to mirror docker.io to return an error if the client actually meant quay.io, the registry just isn't expected to have the same amount of context as a client in regards to the intent of the entire pull process. If the registry chooses to be handle the `ns` parameter and not support it, it is as easy as returning a 404 for unconfigured upstreams.

> How will clients know to be able to check for this capability?

They aren't expected to check for it, but rather be explicitly configured for it. A client will know if it is configured to always use a specific mirror or a mirror for multiple namespaces. I think what you are suggesting here though is the idea of registry discovery. That is a much larger topic that I would still love to see happen, in that feature a client could start with zero knowledge (except of course the domain quay.io, docker.io, etc) and discover registry capabilities and endpoints.

spec.md
```diff
@@ -1260,6 +1276,7 @@ The following parameters SHOULD be specified on the request:
 | `name` | path  | Name of the target repository.                                                              |
 | `n`    | query | Limit the number of entries in each response. It not present, all entries will be returned. |
 | `last` | query | Result set will include values lexically after last.                                        |
+| `ns`   | query | (OPTIONAL) Namespace of repository. SHOULD be set to mirrored host.                         |
```
vbatts: so this SHOULD be `docker.io/` or `quay.io/`

dmcgowan: `docker.io` or `quay.io`, the `/` as a separator is implied by use of paths in the repository name. But can make that explicit.

vbatts:
discussion ensues on the call today.
This sounds like a decent addition, but with a clear use-case for the behavior, and whether a registry implementation MUST support it.

jzelinskie:
pinging @thomasmckay and @kurtismullins who are implementing mirroring on Quay -- they probably have feedback and want to track this thread

RCMariko:
Pulp container plugin team will want to keep an eye on this thread as well. Any feedback @ipanova @dkliban @asmacdo ?

vbatts:
is `ns` already used, so best to continue with that mnemonic? or could
it be something to not collide with the outdated concept that images
would on be named "transport/namespace/name:tag"?

spec.md
kurtismullins: This project is new to me, but I wanted to share some information as this naming convention was a bit confusing to me.
In Quay, `namespace` references a particular User or Organization. For example, given the URI `quay.io/redhat/quay`
- `quay.io` is the hostname/registry
- `redhat` is the namespace
- `quay` is the repository

dmcgowan:
@vbatts I use `ns` or namespace here because namespace is a very generic concept. Certainly as a generic concept it has been used to mean many things. However generally namespace would refer to additional context (such as a prefix) on another name. In the distribution spec case, the name given to the registry would be the part on the URL path, the namespace just gives that additional context to the name. Existing distribution clients today parse the name as you described `<sometransport/host/whatever>/<name given to the registry>`, in which case when given just `<name given to registry>`, the `<sometransport/host/whatever>` would be the namespace of that. You could continue to divide those parts in smaller names in other namespaces, such as the Docker hub does with usernames/reponames, but that is out of scope here. Capturing this in elegant words is kind of tough, recommendations on which parts are unclear or how to make it better are appreciated.

stevvooe (dismissed):
LGTM

I think this is a great addition. It might be a good idea to add a few combinatoric examples of how ns and the the repo name are combined to calculate the upstream and local mirroring location.

vbatts:
@dmcgowan one thing i'm unclear on here is: can i have a single registry mirror that will be usable for more than one remote registry (i.e. remote of `docker.io/...`, `quay.io/...`, etc)

spec.md
majewsky: Yes, the word "namespace" is quite unspecific and its usage here partially conflicts with other appearances of that word within the spec. Particularly, under "Starting an upload", the request `POST /v2/<name>/blobs/uploads/` has a description reading:

> The parameters of this request are the image **namespace** under which the layer will be linked.

That's a different kind of namespace than what is meant here (and even different from how Quay uses the term apparently). This passes into #27 territory and should be cleaned up anyway, but IMO we shouldn't make it worse and use such vague terms in more places.

If the `ns` parameter only has values like `docker.io` or `quay.io`, would it make sense to rename it e.g. `upstream` and explicitly define it as containing the _domain name_ of the upstream registry?

spec.md
```diff
@@ -901,6 +902,20 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
 
 > for more details, see: [compatibility.md](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#backward-compatibility)
 
+### Registry Mirroring
+
+A registry MAY operate as a mirror for an upstream registry to support pull-through caching or proxying of pull operations (such as fetching tags, manifests, or blobs).
```
majewsky: This paragraph was very confusing to me at first because I thought that it refers to a reverse-proxying strategy like the pull-through cache behavior supported by [docker/distribution](https://github.com/docker/distribution) when it's actually a forward proxy. Perhaps that could be clarified?

spec.md
josephschorr: +1 or, alternatively, `upstream-registry` to be specific

spec.md
dmcgowan: The intent is to highlight it can be used for pull-through caching or forward proxying. I don't understand what you mean by docker/distribution behavior being a reverse-proxy strategy though, can you clarify so I can understand how to make this more clear. I ask because I never saw that implementation in docker/distribution as a reverse proxy, it was intending more as a forward proxy but without the capability specified here, that can't be easily done.

spec.md
dmcgowan: > Yes, the word "namespace" is quite unspecific and its usage here partially conflicts with other appearances of that word within the spec. Particularly, under "Starting an upload", the request POST /v2/<name>/blobs/uploads/ has a description reading:

The word "namespace" is intentionally unspecific since it is highly contextual. As pointed out elsewhere, we specify a the `<name>` component in the URL path, the `ns` parameter is further information related to that `<name>`, hence the full name being `<namespace>/<name>`. We should clarify any part of the documentation that suggests that full names are `<host>/<namespace>/<repository>` as that is only referring to a single registry interpretation and this specification is intended to be less opinionated.

>  would it make sense to rename it e.g. upstream and explicitly define it as containing the domain name of the upstream registry

It would be better to clarify how it is applied rather than rename it. The namespace could be the upstream registry, and should be used by mirrors to resolve the host. However, just because this normally represents a domain, does not mean it represents the upstream host. For example, with `docker.io` the upstream host is resolved differently. The intent of this change is to provide the full name used by the client, not to tell the registry host how to use the name. Mirroring is just the use case here, the specification itself is more generic.

spec.md
majewsky: I'm refering to https://docs.docker.com/registry/configuration/#proxy.

> The proxy structure allows a registry to be configured as a pull-through cache to Docker Hub. See mirror for more information. Pushing to a registry configured as a pull-through cache is unsupported.

It's structured as a reverse proxy i.e. it takes in requests that the client makes for the proxy's domain name and rewrites them into requests for a specific backend registry. It does not allow the client to specify which upstream to pull from like a forward proxy would.

The wording in the PR, particularly in line 907 here, says that the proxy operates as a mirror "for an upstream registry". Note the singular.

vbatts:
Notes from today's call:
* (jz) quay does something different for mirroring. This should be called "proxying"
* (dmcg) this is really for client side caching proxy, and is needed.

Please lets find a way to classify this language (whether client or server side). So we can close  out or merge this

jzelinskie:
Reiterating our convo from the meeting:

I actually see a lot of value in adding this query parameter, but removing any connotation that it is the blessed solution for repository mirroring. I think that by including this value, a proxy could implement lots of different behavior for the client that need not be directly related to repository mirroring.

amouat:
I note that harbor uses the term "replication" rather than proxy-cache or mirroring, which I quite like https://goharbor.io/docs/1.10/administration/configuring-replication/

dmcgowan:
Updated section to remove use of the term "mirror". A mirror is just a special case of a proxy or could be a completely different configuration. For example, the most common "mirroring" case which is used by Docker is actually a pull-through cache. Using the more generic terminology here instead of continuing the confusion :)

spec.md
```diff
@@ -840,6 +841,21 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
 
 > for more details, see: [compatibility.md](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#backward-compatibility)
 
+### Registry Proxying
+
+A registry MAY operate as a proxy to another registry to support pull-through caching or proxying of pull operations (such as fetching tags, manifests, or blobs).
```
jzelinskie, line 846: I've broken this down into 3 separate sentences that articulate one clear thought. I think this also makes it more obvious that this feature could be used for more than the examples provided.

>A registry MAY operate as a proxy to another registry to delegate functionality or implement additional functionality.
>An example of delegating functionality is proxying pull operations to another registry.
>An example of adding functionality is implementing a pull through cache of pulls to another registry.

spec.md
```diff
@@ -840,6 +841,21 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
...
+In this case, the `Host` header passed to the proxy registry WILL NOT be the same as the host in the repository name used by the client.
+A `ns` query parameter on pull operations is OPTIONAL, but when used specifies the host in a repository name used by a client.
+This host component in a repository name SHOULD be the registry host a client considers the primary source for a repository, however, a client MAY be configured to use a different host.
+The host in the repository name SHOULD be the first component of the full repository name used by a client.
```
jzelinskie, line 850: I'd put this sentence before the previous one.

spec.md
```diff
@@ -840,6 +841,21 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
...
+This original host component used by the client is referred to as the source host in the API documentation.
+A proxy registry MAY use the `ns` query parameter to resolve an upstream registry host.
+A proxy registry MAY choose to ignore the `ns` query parameter.
+
+A client SHOULD be aware of whether a registry host is being used is a proxy.
```
jzelinskie, line 855: This leads to the question of "how do I do that?"

jzelinskie (dismissed):

spec.md
vbatts: http headers that the client SHOULD log somewhere?

dmcgowan: Sure, the case is usually when http `HOST` does not match the source host in the `ns` query parameter. The client can obviously define their own rules, but can mention this condition as a recommendation based on parts within this specification.

spec.md
dmcgowan: Thanks, agreed, much clearer

jzelinskie:
LGTM

spec.md
```diff
@@ -1019,11 +1037,12 @@ The implementation MAY impose a maximum limit and return a partial set with pagi
...
-|-----------------|--------|----------------------------------------------------------------|
-| `Host`          | header | Standard HTTP Host Header. SHOULD be set to the registry host. |
-| `Authorization` | header | An RFC7235 compliant authorization header.                     |
-| `name`          | path   | Name of the target repository.                                 |
+| Name            | Kind   | Description                                                       |
+|-----------------|--------|-------------------------------------------------------------------|
+| `Host`          | header | Standard HTTP Host Header. SHOULD be set to the registry host.    |
+| `Authorization` | header | An RFC7235 compliant authorization header.                        |
+| `name`          | path   | Name of the target repository.                                    |
+| `ns`            | query  | (OPTIONAL) Namespace of repository. SHOULD be set to source host. |
```
vbatts, line 1045: should we clarify that if `ns` is _not_ provided, there is a default assumption for the source host? or is that implied in the configuration of registry-proxy?

vbatts (dismissed):
LGTM
apart from a curious question

spec.md
```diff
@@ -840,6 +841,23 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
...
+A `ns` query parameter on pull operations is OPTIONAL, but when used specifies the host in a repository name used by a client.
+The host in the repository name SHOULD be the first component of the full repository name used by a client.
+This host component in a repository name SHOULD be the registry host a client considers the primary source for a repository, however, a client MAY be configured to use a different host.
+This original host component used by the client is referred to as the source host in the API documentation.
+A proxy registry MAY use the `ns` query parameter to resolve an upstream registry host.
+A registry MAY choose to ignore the `ns` query parameter.
+
+A client SHOULD be aware of whether a registry host is being used is a proxy, such as when the `ns` query parameter differs from the `Host` header.
+A client SHOULD avoid sending `ns` query parameters to non-proxy registries.
+A client SHOULD NOT unintentionally send authorization credentials for an upstream registry to a proxy registry.
```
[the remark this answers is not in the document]
riking: Alternative wording:

> Clients SHOULD NOT assume that proxy registries can handle authentication credentials issued by an upstream registry. Authentication credentials MUST NOT be sent to different hostnames except when prior knowledge or a configuration signal of authentication support is present.

jdolitsky:
big time rebase needed @dmcgowan

spec.md
```diff
@@ -1019,11 +1037,12 @@ The implementation MAY impose a maximum limit and return a partial set with pagi
 
 The following parameters SHOULD be specified on the request:
 
-| Name            | Kind   | Description                                                    |
-|-----------------|--------|----------------------------------------------------------------|
-| `Host`          | header | Standard HTTP Host Header. SHOULD be set to the registry host. |
-| `Authorization` | header | An RFC7235 compliant authorization header.                     |
-| `name`          | path   | Name of the target repository.                                 |
+| Name            | Kind   | Description                                                       |
```
jdolitsky, line 1040: I think we lost these tables in the spec.md rewrite

spec.md
```diff
@@ -840,6 +841,23 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
 
 > for more details, see: [compatibility.md](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#backward-compatibility)
 
+### Registry Proxying
+
+A registry MAY operate as a proxy to another registry to delegate functionality or implement additional functionality.
+An example of delegating functionality is proxying pull operations to another registry.
+An example of adding functionality is implementing a pull-through cache of pulls to another registry.
+When operating as a proxy, the `Host` header passed to the registry will be the host of the PROXY and NOT the host in the repository name used by the client.
+A `ns` query parameter on pull operations is OPTIONAL, but when used specifies the host in a repository name used by a client.
```
jdolitsky, line 850: this is the key piece - we should document at least this query param

dcarrion87:
Hi team, just checking to see when this is likely to be merged in? Thanks

sudo-bmitch:
@dcarrion87 it's going to be a while. Effort is still continuing on the 1.1 release. After that, we'll revisit this.

dmcgowan:
I don't mind rebasing this after 1.1 to target 1.2. It really is a pretty light change, it has been included by `containerd` since 1.4. The issue on the client is whenever a proxy is used, the client is throwing away information that could be important to the registry. We only include `ns` whenever the proxy host name does not match the requested image host name. I would love to see a registry proxy make use of it even before it hits an official `1.2.0` release.

spec.md
```diff
@@ -2348,13 +2370,14 @@ If the header `Accept-Range: bytes` is returned, range requests can be used to f
...
+| Name            | Kind   | Description                                                       |
+|-----------------|--------|-------------------------------------------------------------------|
+| `Host`          | header | Standard HTTP Host Header.SHOULD be set to the registry host.     |
+| `Authorization` | header | An RFC7235 compliant authorization header.                        |
+| `Range`         | header | HTTP Range header specifying blob chunk.                          |
+| `name`          | path   | Name of the target repository.                                    |
+| `digest`        | path   | Digest of desired blob.                                           |
+| `ns`            | query  | (OPTIONAL) Namespace of repository. SHOULD be set to source host. |
 
 ###### On Success: Partial Content
```
jmzwcn, line 2383: quickly please!

phillebaba:
I just wanted to check in if this proposed change has just stalled or is abandoned? I know that these things take a long time, so it can at times be hard to tell.

I have been working in depth with mirroring OCI artifacts over the last year now. The addition of an `ns` query parameter solves a lot of problems seen in different projects. The major issue right now seems to be that all workarounds are non standard, causing incompatibility. Containerd adds the query parameters whil CRIO does not. Some registries will look for the query parameter while others either only support a single registry or suggest including the root url in the namespace.

I just wanted to add my input that I think that adding this to the spec would be very useful and could potentially help pain points that exist today when mirroring.

If there is any additional work required before merging this, I would be happy to help.

brandond:
I've been watching this as well, and would like to voice my support for getting this officially included in the spec.

darzanebor:
+1

BenTheElder:
Operating https://registry.k8s.io  we see a TON of requests that are obviously from pull-through caches/mirrors like artifactory for images that we don't host. If this API was standardized we could eventually start requesting that these tools switch to ns-aware pulls and stop spamming us with requests for images we don't provide. 

While not the top of our concerns, this feels wasteful and generates a lot of log noise, and isn't free.

The current state where pull-through mirroring pretty much has to request all upstreams feels bad, what concerns remain with this proposal?

sudo-bmitch:
> Operating https://registry.k8s.io  we see a TON of requests that are obviously from pull-through caches/mirrors like artifactory for images that we don't host. If this API was standardized we could eventually start requesting that these tools switch to ns-aware pulls and stop spamming us with requests for images we don't provide. 

I'd guess this is from users thinking they have setup a proxy, and docker automatically falling back when the proxy pull fails. This is made worse by docker only supporting a proxy to Docker Hub. The result is the proxy does nothing, never used to pull k8s images (since the client would pull directly) and never finding a match (since the proxy isn't pointing to Hub) for the Hub pulls.

> The current state where pull-through mirroring pretty much has to request all upstreams feels bad, what concerns remain with this proposal?

Are there proxies doing this today? The few I've used are configured for a single upstream registry, at least for a given namespace or port.

I worry we'll all agree this is a good change, but the key players in the space won't implement it. It would be good to see this being actively used by both clients and proxies.

We also need to consider existing clients and proxies in various combinations. E.g. if a proxy for a single registry is working with older clients and gets upgraded to use the headers, those older clients still need to go to the original registry. Similarly, newer clients configured with an older proxy should not pull content from the wrong upstream registry, since that would expose them to attacks by dependency confusion and namespace squatters.

BenTheElder:
> I'd guess this is from users thinking they have setup a proxy, and docker automatically falling back when the proxy pull fails. This is made worse by docker only supporting a proxy to Docker Hub. The result is the proxy does nothing, never used to pull k8s images (since the client would pull directly) and never finding a match (since the proxy isn't pointing to Hub) for the Hub pulls.

The issue where docker only supports configuring a mirror for dockerhub is also pretty unfortunate but not what I'm referring to, the headers are for agents like artifactory (and other similar tools).

My understanding is that users setup a central catch-all pull-through cache and have no way to disambiguate which upstream the pulls were meant for, so we get pulls for ... a lot of images clearly not associated with us.

This seems to be common with self hosted options, versus something like ECR/AR remote repositories. It's understandable that adding a distinct IP / hostname for every proxied host is heavy handed, and I don't think most users even think to use a port other than 5000.

> Are there proxies doing this today? The few I've used are configured for a single upstream registry, at least for a given namespace or port.

AFAICT, yes, for example:

https://jfrog.com/help/r/jfrog-artifactory-documentation/virtual-docker-repositories

"This allows you to access images that are hosted locally on local Docker repositories, as well as remote images that are proxied by remote Docker repositories, and access all of them from a single URL defined for the virtual repository. Using virtual repositories can be very useful since users will continue to work with the virtual repository while the admin can manage the included repositories, replace the default deployment target and those changes will be transparent to the users. [...]"

> I worry we'll all agree this is a good change, but the key players in the space won't implement it. It would be good to see this being actively used by both clients and proxies.

This seems a bit circular? If we don't document the spec for this optional feature then most projects are not going to enable support (e.g. https://github.com/distribution/distribution/pull/3864#issuecomment-1974748544)

> We also need to consider existing clients and proxies in various combinations. E.g. if a proxy for a single registry is working with older clients and gets upgraded to use the headers, those older clients still need to go to the original registry. Similarly, newer clients configured with an older proxy should not pull content from the wrong upstream registry, since that would expose them to attacks by dependency confusion and namespace squatters.

That's true, but a proxy can behave more intelligently when the namespace parameter is present and over time more clients will have it, there are various references to waiting for this to be standard. The proxies can fall back to existing behavior such as querying all upstreams when the header is missing.

Containerd alone represents a lot of clients that already have this header in place. Were it to become standard and proxies begin supporting it there's actually a pretty large client base sending it already, but there's resistance to supporting it more widely referring back to waiting for this PR.

sudo-bmitch:
> The issue where docker only supports configuring a mirror for dockerhub is also pretty unfortunate but not what I'm referring to, the headers are for agents like artifactory (and other similar tools).

I'm not sure how you'd differentiate the scenarios from the k8s registry. It would just look like a bunch of failed image pulls to that registry with either scenario.

> > Are there proxies doing this today? The few I've used are configured for a single upstream registry, at least for a given namespace or port.
> 
> AFAICT, yes, for example:
> 
> https://jfrog.com/help/r/jfrog-artifactory-documentation/virtual-docker-repositories

Ugh, someone should open a dependency confusion CVE against them if that's really what's happening.

> > I worry we'll all agree this is a good change, but the key players in the space won't implement it. It would be good to see this being actively used by both clients and proxies.
> 
> This seems a bit circular? If we don't document the spec for this optional feature then most projects are not going to enable support (e.g. https://github.com/distribution/distribution/pull/3864#issuecomment-1974748544)

We've always standardized things that were already implemented somewhere to show the feasibility first. You can't untag a standard, and implementations are effectively the CI for a standard. If the distribution project doesn't want to add it, or maintain a feature branch, then we'd need another project.

> > We also need to consider existing clients and proxies in various combinations. E.g. if a proxy for a single registry is working with older clients and gets upgraded to use the headers, those older clients still need to go to the original registry. Similarly, newer clients configured with an older proxy should not pull content from the wrong upstream registry, since that would expose them to attacks by dependency confusion and namespace squatters.
> 
> That's true, but a proxy can behave more intelligently when the namespace parameter is present and over time more clients will have it, there are various references to waiting for this to be standard. The proxies can fall back to existing behavior such as querying all upstreams when the header is missing.

Registries falling back when the header isn't seen is probably easy enough. That needs to be to a single registry to avoid a dependency confusion attack.

The fallback for new clients talking to older registries is less clear. The API needs to ensure new clients either know to fallback when talking to older proxies, or that the API doesn't exist on old proxies to avoid the security risk of sending requests to the wrong upstream registry.

> Containerd alone represents a lot of clients that already have this header in place. 

With Docker updating that integration, the single proxy setting to Hub will hopefully stop being an issue in the future.

phillebaba:
If you need an example of the `ns` query parameter being used in the wild here is one. I am currently using it in Spegel do enable tag resolving in a simple way.  This is less of an issue for me as I only support Containerd which has been adding this parameter to all of its image pulling request for a few years now. This has been working really well as from a user perspective it becomes  a seamless experience.

https://github.com/spegel-org/spegel/blob/ecfdd57cb2a22513c5785cce89e5c8eb6771dbf3/pkg/registry/registry.go#L180-L185

The other solution that I have seen used out in the wild is to add a header to the mirror configuration so that the receiving registry is aware of the original registry the image originates from. This is how Drangonfly solves the problem for example. Which means that they would not be able to support wild card mirror configuration.

https://d7y.io/docs/next/operations/integrations/container-runtime/containerd/#multiple-registries

Right now every registry out there seems to come up with their own standard for how to deal with mirroring multiple namespaces. A common practices is to to join the original image name reference with the mirror registry. For example `docker.io/library/ubuntu:20.04` would become `example.com/docker.io/library/ubuntu:20.04`. While it may work it creates another set of challenges forcing users to modify image reference and creating incompatibility with other tools that do not understand the logic implemented in the name.

sudo-bmitch:
Can this be used more generically? Not just for image pulls from a proxy, but also for tag listings, referrers, possibly even a push-through proxy? Is there a way users can directly reference content on a proxy with a namespace parameter, e.g. an added field in the image reference like `proxy.example.org/some/image:some-tag[registry.example.com/some/image]`? I'm considering how this could be extended to accessing content in an OCI Layout were an export has multiple images with the same tag but different source repositories.

ordovicia:
As a real-world example [CIRC, introduced in this KubeCon session](https://sched.co/1x708), also uses the `ns` query parameter and it really works well.
CIRC is a cache server for container images, and we configure containerd to send all image pull requests to CIRC.
As containerd adds this parameter to HTTP requests, CIRC is able to identify the origin registry, enabling transparent proxying for arbitrary registries.
I would like to see `ns` query parameter being defined in the distribution spec, and other container runtimes implement it.

brandond:
Yeah I think it's safe to say that ns has become a de facto standard despite this PR stalling out.

sudo-bmitch:
If the registry ignores a `ns` value, what are the allowed responses? Should the proxy notify the client about which upstream registry it used? I'm worried about dependency confusion attacks where a client thinks it is pulling from one registry, but the proxy responds with content from another location. A header from the proxy, indicating the upstream registry, would allow the client to cross reference the expected upstream and reject the proxy response on a mismatch.

sudo-bmitch:
> > I'm worried about dependency confusion attacks where a client thinks it is pulling from one registry, but the proxy responds with content from another location
> 
> This is not a problem that needs to be solved here. Tag mutability, and figuring out what a tag points at when pulling from different registries or mirrors, is a known issue. If this matters to you, the correct solution is to use digests instead of tags. Beyond that, it doesn't matter where the content comes from. Allowing the client to make decisions based on where a registry reports that it is getting its content from violates separation of concerns.
> 
> Pull a digest instead of a tag, and you're guaranteed to get what you want. Any hacks around "trusting" content because the server tells you it got it from a specific location, are insecure by design.

I completely agree to using digests and signing for security, even when no proxies are involved. But we have a scenario where it's less the client trusting the proxy, and more the proxy telling the client "you asked for X and I'm choosing to give you Y instead". Or we could say proxies should not do that. Or we can keep the current language that says a proxy is within the spec to return different content than what the client requested without any notification to the client. Of the three options, another header to give the client some feedback seems the least intrusive and most flexible to implementations.

sudo-bmitch (approved):
Following up on the Thursday call discussion, I added a commit with the `OCI-Namespace` header. If that's blocking other maintainers from approving, I can split that into a separate PR.

mikebrow (commented):
couple nits

dmcgowan:
Updated, not intentionally removed, just never had it in my local branch when rebased

jdolitsky (approved):

mikebrow (approved):
LGTM

PR #39, opened 2018-12-21
Import rest of tests from kinvolk/ocicert

dongsupark:
This PR basically combines 2 original PRs into one: https://github.com/kinvolk/ocicert/pull/2 and https://github.com/kinvolk/ocicert/pull/5

The PR adds more tests such as UploadLayer, PushManifest, PullManifest, PullLayer, DeleteLayer, DeleteManifest.
Review comments were partly addressed. There are still open issues like [this](https://github.com/kinvolk/ocicert/pull/5#discussion_r242939621).

If there's any comment, please comment to this PR. I can update it.

PR #67, opened 2019-07-31
Update blob patch status code to correct code

dmcgowan:
The PATCH blob upload endpoint returns a 202 Accepted when a chunk
is accepted rather than a 204. The 204 is returned by the PUT
endpoint after completion of an upload is successful. The status
code in the detail section must match the protocol as described
in the description for upload patching. It is stated at the beginning
of the detail section that when the detail does not match, the
description above it is correct.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -3740,7 +3740,7 @@ The following parameters SHOULD be specified on the request:
 ###### On Success: Data Accepted
 
 ```HTTP
-204 No Content
+202 Accepted
 Location: /v2/<name>/blobs/uploads/<uuid>
 Range: 0-<offset>
 Content-Length: 0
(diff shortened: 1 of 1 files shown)


vbatts:
LGTM

jzelinskie:
Are we at a point where we could write a test for this?

vbatts:
the tests in place do some, but not all.
https://github.com/opencontainers/distribution-spec/pull/39 was never merged, which went further

jonjohnsonjr:
> The 204 is returned by the PUT endpoint after completion of an upload is successful.

Is this actually true? AFAICT registries actually return a 201 here, and the spec has a similar discrepancy between [the prose](https://github.com/opencontainers/distribution-spec/blame/75485b8b99670c52a5d462995aefb532068508a2/spec.md#L547) and [the detail](https://github.com/opencontainers/distribution-spec/blame/75485b8b99670c52a5d462995aefb532068508a2/spec.md#L4228).

dmcgowan:
@jonjohnsonjr you are right, that one has a mismatch as well. It seems the only 204 should be on the upload progress. Delete is also wrong. I'll open up a separate PR for those so the status codes can be evaluated separately.  Ideally we can get rid of the disclaimer at the top of the detail section stating that it may be wrong :)

PR #75, opened 2019-12-20
spec: clarify digest algorithm

### LINKED ISSUE #63: Content-Digest: digest algorithm requirement is ambiguous
In https://github.com/opencontainers/distribution-spec/blob/master/spec.md#content-digests, the following statement fill a bit ambiguous to me:
> While the algorithm does allow one to implement a wide variety of algorithms, compliant implementations SHOULD use sha256

The ambiguity is on the _SHOULD_ word: if my registry uses sha512 for digesting contents (and thus for OCI descriptors), is it still compliant to the spec ?

More importantly, as a consumer of the registry API, can I rely on the fact that all OCI-compliant registries use sha256 for their descriptors ?
If it is not the case, it is a very serious limitation for interoperability: it means that I cant update a manifest with its dependencies in OCI Image layout from registry-a and be sure I can upload it as-is in registry-b. If the 2 registries use different digest algorithms for their descriptors, I need to modify all the descriptors in my manifests when doing a copy.
It also means that any digital signature I have for trusting a given manifest from registry-a can't be used to validate the same manifest once copied on registry-b (because my descriptors would have been rewritten).

vbatts:
While other algorithm's can be used, signal the portability issues.

Fixes #63

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -240,9 +240,10 @@ Some examples of _digests_ include the following:
 
 | digest                                                                  | description                |
 |-------------------------------------------------------------------------|----------------------------|
-| sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b | Common sha256 based digest |
+| `sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b` | Common sha256 based digest format |
 
-While the _algorithm_ does allow one to implement a wide variety of algorithms, compliant implementations SHOULD use sha256.
+Compliant implementations SHOULD use _algorithm_ of sha256.
+The _algorithm_ does allow one to implement a wide variety of algorithms, though for portability sake it is best to keep this consistent because a translation of hashing algorithm used would produce a new object and new documents that reference it.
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -240,9 +240,10 @@ Some examples of _digests_ include the following:
 
 | digest                                                                  | description                |
 |-------------------------------------------------------------------------|----------------------------|
-| sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b | Common sha256 based digest |
+| `sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b` | Common sha256 based digest format |
 
-While the _algorithm_ does allow one to implement a wide variety of algorithms, compliant implementations SHOULD use sha256.
+Compliant implementations SHOULD use _algorithm_ of sha256 (until the ruled sufficiently broken).
```
dmcgowan: reword or remove the parenthetical statement

vbatts: removed. Since this will be captured in a version, it will stand as that version. No need to allude to removal.

jzelinskie:
LGTM

PR #76, opened 2019-12-20
spec: session_id is clearer than UUID

### LINKED ISSUE #13: spec: Consider removing references to UUID
The spec constantly refers to UUID that reference a given upload e.g. the `Docker-Upload-UUID` header.

However, there's no real reason this needs to be a UUID, it's just a reference number or ID (as noted in the spec itself "While the uuid parameter may be an actual UUID, this proposal imposes no constraints on the format and clients should never impose any"). Could it be changed to just `id` or `refid` or some such?

vbatts:
the behavior previously described as `UUID` is really a session ID.
For implementations with existing behavior this is a nop, but going
forward the header `Blob-Upload-Session-ID` is more clear than
`Docker-Upload-UUID`.

Fixes #13

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -425,7 +425,7 @@ If the POST request is successful, a `202 Accepted` response will be returned wi
 
 ```HTTP
 202 Accepted
-Location: /v2/<name>/blobs/uploads/<uuid>
+Location: /v2/<name>/blobs/uploads/<session_id>
 Range: bytes=0-<offset>
 Content-Length: 0
 ```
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -3425,15 +3425,15 @@ The following parameters SHOULD be specified on the request:
 | `Host`          | header | Standard HTTP Host Header.SHOULD be set to the registry host.                                 |
 | `Authorization` | header | An RFC7235 compliant authorization header.                                                    |
 | `name`          | path   | Name of the target repository.                                                                |
-| `uuid`          | path   | A uuid identifying the upload.This field can accept characters that match `[a-zA-Z0-9-_.=]+`. |
+| `session_id`          | path   | A unique string identifying session of the partiuclar upload.This field can accept characters that match `[a-zA-Z0-9-_.=]+`. |
```
dmcgowan: particular*

spec.md
```diff
@@ -3898,16 +3898,16 @@ The following parameters SHOULD be specified on the request:
...
 | `Content-Length` | header | Length of the chunk being uploaded, corresponding the length of the request body.                                                                                                                                 |
 | `name`           | path   | Name of the target repository.                                                                                                                                                                                    |
-| `uuid`           | path   | A uuid identifying the upload.This field can accept characters that match `[a-zA-Z0-9-_.=]+`.                                                                                                                     |
+| `session_id`           | path   | A unique string identifying session of the partiuclar upload.This field can accept characters that match `[a-zA-Z0-9-_.=]+`.                                                                                                                     |
```
dmcgowan: particular*

spec.md
vbatts: got it

jzelinskie:
LGTM

Would it make sense to also prefix any headers with `X-`? e.g. `X-Blob-Upload-Session-ID`

jonjohnsonjr:
> Would it make sense to also prefix any headers with X-?

Someone recently informed me this practice is deprecated: https://tools.ietf.org/html/rfc6648.html

vbatts:
> > Would it make sense to also prefix any headers with X-?
> 
> Someone recently informed me this practice is deprecated: https://tools.ietf.org/html/rfc6648.html

that's what i thought too, but you found it first

vbatts:
@jonjohnsonjr thought on this PR? You had GCR thoughts on #13

jonjohnsonjr:
LGTM, this part covers the behavior we expect from clients:

> Though the URI format (`/v2/<name>/blobs/uploads/<session_id>`) for the `Location` header is specified, clients SHOULD treat it as an opaque url and SHOULD never try to assemble it.

vbatts:
cool cool

jzelinskie:
@jonjohnsonjr @vbatts oh nice! I'm super late to seeing that RFC.

Issue #79, opened 2020-01-02
clarification: "Location" header bevahior?


rchincha:
As per spec, "Location" header is typically of the form:
Location: /v2/<name>/blobs/uploads/<session_id>

Also, 
```
Though the URI format (/v2/<name>/blobs/uploads/<session_id>) for the Location header is specified, clients SHOULD treat it as an opaque url and SHOULD never try to assemble it.
```

However, docker distribution appears to return a fully qualified "Location" header value of the form:
https://<server_ip>:<server_port>/v2/<name>/blobs/uploads/<session_id>

This behavior needs further clarification. What should be a compliant client's expectation from the server? In one case, it needs to prepend the Location header with the (scheme,address,port) and in the other, not.

jonjohnsonjr:
I think we can just defer to [RFC 7231](https://tools.ietf.org/html/rfc7231#section-7.1.2):

> The field value consists of a single URI-reference.  When it has the
   form of a relative reference ([[RFC3986], Section 4.2](https://tools.ietf.org/html/rfc3986#section-4.2)), the final
   value is computed by resolving it against the effective request URI
   ([[RFC3986], Section 5](https://tools.ietf.org/html/rfc3986#section-5)).

rchincha:
So this would mean the client needs to distinguish between the two Location formats, which is fine. Should this be called out in the dist-spec?

jzelinskie:
Absolutely. We should add a link to the RFC in all the tables that document the headers for endpoints -- some of them are even empty for Location.

jdolitsky:
Resolved by #206

PR #80, opened 2020-01-06
spec: `Location` can be absolute or relative

rchincha:
From the spec it is not clear how the `Location` header must be set.
This comment clarifies the behavior.

Fixes issue #79

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -387,6 +387,8 @@ Uploads are started with a POST request which returns a url that can be used to 
 
 The `Location` header will be used to communicate the upload location after each request.
 While it won't change in the this specification, clients SHOULD use the most recent value returned by the API.
+`Location` header value returned MUST either be absolute or relative as described in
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -387,6 +387,8 @@ Uploads are started with a POST request which returns a url that can be used to
 
 The `Location` header will be used to communicate the upload location after each request.
 While it won't change in the this specification, clients SHOULD use the most recent value returned by the API.
+Note that the `Location` header value returned may either be absolute or relative as described in
```
dmcgowan: The "Note that" doesn't seem necessary, try "The `Location` header value MAY be either an absolute or relative URI as described in"

rchincha: Pushed an update per comments. MUST instead of MAY makes more sense here?

vbatts:
LGTM

PR #90, opened 2020-01-29
Conformance: Add CI, new tests, more report detail

### LINKED ISSUE #83: conformance: Add CI for test source
Add Travis job to do things such as:

- Run golangci-lint
- Make sure that the test binary builds
- Run against local registry w/ known compliance (zot?)

See https://github.com/opencontainers/distribution-spec/pull/82#issuecomment-576438419

### LINKED ISSUE #84: conformance: Add test summary at the top of HTML report
Should show a summary of whether or not tests passed, how many passed/failed, and the configuration used to run the tests (which registry URL etc)

### LINKED ISSUE #87: conformance: test for error codes
On failed request (4xx), check for JSON response body.  In the body, the "code" key should be all-caps with underscores, taken from this list in the spec:

https://github.com/opencontainers/distribution-spec/blob/master/spec.md#errors-2

### LINKED ISSUE #91: conformance: Test binary should contain info about git commit etc.
`conformance.test` should be able to be built with arguments like git commit etc, so that in the test report it is clear which version of the code built it

pmengelbert:
Add new tests and generate more detailed report

Implement test for monolithic POST blob upload
The specification dictates:
"POST Initiate Blob Upload: Initiate a resumable blob upload. If
successful, an upload location will be provided to complete the upload.
Optionally, if the digest parameter is present, the request body will be
used to complete the upload in a single request."
https://github.com/opencontainers/distribution-spec/blob/master/spec.md#initiate-blob-upload

Implement test for manifest deletion by tag
The specification dictates:
"DELETE Manifest:
Delete the manifest identified by name and reference. Note that a
manifest can only be deleted by digest."

Modify check for JSON body
There was an error in the way improperly-parsed JSON was being handled.
This commit resolves that issue.

Add test for out-of-order chunked blob upload
As per the spec:
"There is no enforcement on layer chunk splits other than that the
server MUST receive them in order. The server MAY enforce a minimum
chunk size. If the server cannot accept the chunk, a 416 Requested
Range Not Satisfiable response will be returned and will include a
Range header indicating the current status:"

Resolves #87

Improve HTML report
HTML report now shows a summary of passes, failures, environment
variables and timestamps.

Resolves #84

Include commit hash of version used to compile binary in HTML report.
Resolves #91

Update travis and makefile to run linter against conformance tests
Resolves #83

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 EPOCH_TEST_COMMIT	:= 91d6d8466e68f1efff7977b63ad6f48e72245e05
+CURRENT_COMMIT	:= $(shell git log --format="%H" -n 1)
 
-DOCKER		?= $(shell command -v docker 2>/dev/null)
-PANDOC		?= $(shell command -v pandoc 2>/dev/null)
+DOCKER	?= $(shell command -v docker 2>/dev/null)
+PANDOC	?= $(shell command -v pandoc 2>/dev/null)
+GOLANGCILINT	?= $(shell command -v golangcli-lint 2>/dev/null)
 
 OUTPUT_DIRNAME	?= output/
 DOC_FILENAME	?= oci-distribution-spec
@@ -19,6 +21,26 @@ ifeq "$(strip $(PANDOC))" ''
 			$(PANDOC_CONTAINER)
 		PANDOC_SRC := /input/
 		PANDOC_DST := /
+	endif
+endif
+
+# Using older version of golangci-lint
+# see https://github.com/golangci/golangci-lint/issues/825#issuecomment-553210121
+GOLANGCILINT_CONTAINER ?= docker.io/golangci/golangci-lint:v1.17.1
+ifeq "$(strip $(GOLANGCILINT))" ''
+	ifneq "$(strip $(DOCKER))" ''
+		GOLANGCILINT = $(DOCKER) run \
+			-it \
+			--rm \
+			-v $(shell pwd)/:/input:ro \
+			-e GOCACHE=/tmp/.cache \
+			-e GO111MODULE=on \
+			--entrypoint /bin/bash \
+			-u $(shell id -u) \
+			--workdir /input \
+			$(GOLANGCILINT_CONTAINER)
+		GOLANGCILINT_SRC := /input/
+		GOLANGCILINT_DST := /
 	endif
 endif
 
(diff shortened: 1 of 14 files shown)


conformance/03_blob_upload_monolithic_test.go
```diff
@@ -18,6 +18,25 @@ var test03BlobUploadMonolithic = func() {
...
+				SetQueryParam("digest", configDigest).
+				SetBody(configContent)
+			resp, err := client.Do(req)
+			Expect(err).To(BeNil())
+			Expect(resp.StatusCode()).To(Equal(http.StatusCreated))
+			lastResponse = resp
+		})
+
+		g.Specify("GET request to blob URL should yield 200", func() {
+			req := client.NewRequest(reggie.GET, lastResponse.GetRelativeLocation())
```
jdolitsky: Do we need to use `lastResponse.GetRelativeLocation()` here? This seems to be the session URL. I think this should just be a normal blob GET no?

pmengelbert: It's a convenience.  The Location header after an upload contains the location of the blob, though it can be accessed by its digest as well.  I changed it to the latter to make it look more like the spec.

pmengelbert:
Our revised report, with the summary at the top:
new-report

pmengelbert:
Here's a sample of what the report looks like:

new-report

conformance/03_blob_upload_monolithic_test.go
```diff
@@ -18,6 +18,25 @@ var test03BlobUploadMonolithic = func() {
...
 
+		g.Specify("POST request with digest and blob should yield a 201", func() {
+			req := client.NewRequest(reggie.POST, "/v2/<name>/blobs/uploads/").
+				SetHeader("Content-Length", configContentLength).
+				SetHeader("Content-Type", "application/octet-stream").
+				SetQueryParam("digest", configDigest).
+				SetBody(configContent)
+			resp, err := client.Do(req)
+			Expect(err).To(BeNil())
+			Expect(resp.StatusCode()).To(Equal(http.StatusCreated))
```
rchincha: https://github.com/opencontainers/distribution-spec/blob/master/spec.md#on-success-created-1

Needs a Location and Session-ID check (whatever this check means). I would definitely add at least the Location check.

pmengelbert: Thanks for the heads up.  We've added a location check anywhere there's a POST request or we are expecting a 201 (Created) response.

rchincha:
The conformance test harness now looks fairly complete. Nice work!

Makefile
```diff
@@ -55,9 +77,23 @@ $(OUTPUT_DIRNAME)/$(DOC_FILENAME).html: header.html $(DOC_FILES) $(FIGURE_FILES)
 endif
 
 header.html: .tool/genheader.go specs-go/version.go
+	# create a symlink to opencontainers if we are running in a fork
```
dmcgowan: This isn't necessary to include in the Makefile, this is not something that is generally supported or encouraged by packages

pmengelbert: Taken care of

PR #95, opened 2020-02-12
Fix typo in pagination section

### LINKED ISSUE #94: Minor error in pagination section
Under https://github.com/opencontainers/distribution-spec/blob/master/spec.md#pagination

The text reads "If the value of n is 2, v1 and v1 will be returned", presumably this should be "If the value of n is 2, v1 and *v2* will be returned".

amouat:
"_v1_ and _v1_ will be returned" -> "_v1_ and _v2_ will be returned"

Fixes #94.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -802,7 +802,7 @@ v3
 v4
 ```
 
-If the value of `n` is 2, _v1_ and _v1_ will be returned on the first response.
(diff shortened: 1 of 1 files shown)


amouat:
Looks like I need to check this out to sign it properly. Should only take a minute...

vbatts:
LGTM

PR #96, opened 2020-02-12
Fix misspelling of octet

dmcgowan:
Get rid of that extra `c`


vbatts:
`application/ocelot-stream` :cat2:

vbatts:
LGTM

PR #104, opened 2020-02-28
Move Dockerfile to conformance directory

### LINKED ISSUE #99: conformance: move Dockerfile to conformance directory
See https://github.com/opencontainers/distribution-spec/pull/98#pullrequestreview-357784248

jdolitsky:
In order to prevent confusion about the purpose of the Dockerfile,
move it into conformance directory where it is more appropriate.

Also remove dependency on Makefile, and instead using build args
to pass along the test version that is visible in report.html.

Resolves #99.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -8,4 +8,4 @@ inputs: {}
 outputs: {}
 runs:
   using: docker
-  image: Dockerfile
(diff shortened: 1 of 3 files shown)


jzelinskie:
LGTM

PR #105, opened 2020-02-28
Allow range of response codes for manifest delete

### LINKED ISSUE #102: conformance: DELETE request to manifest URL (tag) MUST fail
I think [this test](https://github.com/opencontainers/distribution-spec/blob/2ce66de1cd5234a4b5e6721c8439c5fc435d977f/conformance/08_manifest_delete_test.go#L14) is a little over-zealous.

From the spec:

> Delete the manifest identified by name and reference. Note that a manifest can only be deleted by digest.

This seems to me more a description of the registry's original behavior (really, a limitation IMO) and how clients should expect to interact with registries. It doesn't specifically require registries to fail if you delete by tag.

GCR will never be able to pass these conformance tests, because it has some semantics outside of this specification and treats tag references differently from digest references. (I think quay has the opposite behavior from GCR, which is fun 😄 )

To explain a bit further: GCR views tags as a separate entity (a la git tags) as just pointers into the dag. You can delete a _tag_ separately from deleting a _manifest_ by just doing `DELELTE /v2/.../manifests/<tag>`. In fact, GCR _requires_ you to delete tags before deleting an image by digest in order to prevent you from accidentally deleting a tagged image.

Given that this behavior is a super-set of the registry spec, I think it's a reasonable thing for GCR to do, but the conformance tests treat this behavior as a failure.

WDYT? Am I just biased here?

pmengelbert:
Resolves #102.  The revised test checks for one of three
responses: 202 Accepted, 400 Bad Request, or 405 Method Not Allowed.  If
the response is 400 Bad Request, an additional check is triggered to
verify an OCI-Conformant code is returned, in this case "UNSUPPORTED".

Signed-off-by: Peter Engelbert <pmengelbert@gmail.com>

### DIFF
--- a/conformance/08_manifest_delete_test.go
+++ b/conformance/08_manifest_delete_test.go
@@ -11,35 +11,50 @@ import (
 
 var test08ManifestDelete = func() {
 	g.Context("Manifest Delete", func() {
-		g.Specify("DELETE request to manifest URL (tag) MUST fail", func() {
+		g.Specify("DELETE request to manifest should return 202, 400, or 405", func() {
 			req := client.NewRequest(reggie.DELETE, "/v2/<name>/manifests/<reference>",
 				reggie.WithReference(firstTag))
 			resp, err := client.Do(req)
 			Expect(err).To(BeNil())
-			Expect(resp.StatusCode()).To(BeNumerically(">=", 400))
-			Expect(resp.StatusCode()).To(BeNumerically("<", 500))
+			Expect(resp.StatusCode()).To(SatisfyAny(
+				Equal(http.StatusBadRequest),
+				Equal(http.StatusMethodNotAllowed),
+				Equal(http.StatusAccepted)))
+			if resp.StatusCode() == http.StatusBadRequest {
+				errorResponses, err := resp.Errors()
+				Expect(err).To(BeNil())
+				Expect(errorResponses).ToNot(BeEmpty())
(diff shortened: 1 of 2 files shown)


conformance/08_manifest_delete_test.go
```diff
@@ -11,28 +11,33 @@ import (
...
-		})
-
-		g.Specify("DELETE request to manifest URL (digest) should yield 202 response", func() {
+		g.Specify("DELETE request to manifest URL (digest) should yield 202 response unless already deleted", func() {
 			req := client.NewRequest(reggie.DELETE, "/v2/<name>/manifests/<digest>", reggie.WithDigest(manifestDigest))
 			resp, err := client.Do(req)
 			Expect(err).To(BeNil())
-			Expect(resp.StatusCode()).To(Equal(http.StatusAccepted))
+			// In the case that the previous request was accepted, this may or may not fail (which is ok)
+			if !deleteByTagAccepted {
```
dmcgowan: I think this is trying to codify too much logic here. Whether the first succeeded shouldn't affect this one. Once we clarify delete as optional, then can just do a similar check as above.

pmengelbert: @dmcgowan thanks; I just pushed a version that doesn't attempt to codify that logic.  The test in question now allows the same range of response codes, plus 404 to cover the case that delete succeeded.

vbatts:
LGTM

Issue #106, opened 2020-03-03
spec: Clarifications about tags listing pagination


jonjohnsonjr:
When we dropped catalog from the spec, we lost [this](https://github.com/opencontainers/distribution-spec/commit/b4e98332eccaf51d1974c9136caf67317bd99104#diff-958e7270f96f5407d7d980f500805b1bL4682) line:

> The implementation MAY impose a maximum limit and return a partial set with pagination links.

Which is under the non-paginated Catalog Fetch request, i.e. the request did not include an `n` parameter.

Some questions:

1. Should this have applied to tag listing as well as catalog?
2. Should this also apply to paginated requests?
3. If the answer is no for 1 or 2, what should registries do if they don't want to return the complete set?

Currently, ECR will return an error if you send a request where n > 1000. This seems somewhat reasonable, but I think it's easier for clients to just handle the partial set + Link header response transparently, as was expected for non-paginated catalog requests.

For a lot of registries, if you don't set `n` explicitly, they will default to some reasonable but low number. This can dramatically increase latency if you want to actually fetch the entire set of tags for a large repository.

There's currently no way to specify "give me everything you have, as fast as you are willing to give to me". For clients that means figuring out, empirically, what the maximum value for `n` should be -- maybe even on a per-registry basis, which seems like something the spec is trying to avoid.

Seems like we could do two things:

1. Clarify that registries MUST return the entire set for non-paginated requests.
2. Clarify that registries MAY return a partial set with pagination links for both non-paginated and paginated requests.

Any thoughts?

samuelkarp:
Put me firmly in the camp of option 2.

Option 1 requires a registry implementer to support an unbounded API.  Unbounded API operations can seem easier at first, but as the data set increases in size, timeouts and response sizes can become a problem.  Providing a paginated API sets expectations that clients will need to retrieve all pages to see results and provides a registry implementer the ability to make assurances around both the maximum size of a given response and the maximum time a call is expected to take.

PR #107, opened 2020-03-03
Clarify tags listing behavior

jonjohnsonjr:
Fixes https://github.com/opencontainers/distribution-spec/issues/106

Registries MAY return a partial set for paginated and non-paginated tag
listing requests.

Also fix a typo.

This implements my second proposed option.


vbatts:
LGTM

PR #111, opened 2020-03-11
extension proposal

### LINKED ISSUE #74: add "feature flags", and API to expose
Rather than incrementing the distribution spec version as certain optional features are added, this API and mapped values of "feature flags" will be used by registries to indicate whether new features are implemented. (Like PEP or similar extensions)

vbatts:
:dancers: 

The goal of this PR is to enable API extensions to the distribution-spec (registry API), so that implementations may experiment, and then once ironed out can propose the API paths be documented canonically here in the OCI distribution-spec.

There will be a place in this spec repo for official extensions to live once ratified. But importantly, it should provide a method for registry/client implementors to iterate on extensions before surfacing them to be solidified in the specification.

The challenges faced are:
* how to span registry-wide vs specific repo/app contexts
* honor ACL handling
* not pollute every existing endpoint

Fixes #74


vbatts:
@opencontainers/distribution-spec-maintainers PTAL
@josephschorr I would imagine the pub-sub to be under extensions here
@justincormack @jonjohnsonjr  I would imagine a signature extension here, so that a signature payload on a digest could be pushed to a registry, then the signature extension allows discovering the digests of signatures associated

ext/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Name
+
+Extension names MUST be unique.
+Names SHOULD include a version.
+
+Each extension's endpoints will be nested below its name.
+
+```HTTP
+    GET /v2/ext/0/...
```
mikebrow: the `/v2/ext/` string, in view of the distribution spec, suggests we are reserving `ext` as a logical registry name to be used for registry extension apis. We may need something like `/v3-ext/0' to make that happen?

ext/README.md
```diff
@@ -0,0 +1,49 @@
...
+| `$name`  | Definition | Summary |
+|:--:|:--:|:--:|
+| 0 | [ext-0](./ext-0.md) | Base definition of discovering extensions on registry server |
+|  |  |  |
+
+
+## Name
+
+Extension names MUST be unique.
+Names SHOULD include a version.
```
mikebrow: What model for version naming should be used

ext/ext-0.md
```diff
@@ -0,0 +1,14 @@
+# ext-0 -- Index of Distribution Extensions
+
```
mikebrow: todo metadata template...

mikebrow (commented):
some comments

ext/README.md
dmcgowan: For `registry` level operations under `/v2` in the past we prepended `_`. I think it is best to keep that convention here to not interfere with repository naming.

We could have a wider discussion on the role of `/v2` and the potential to rev that or just make that not required. Personally I think the `/v2` should remain an implementation detail and avoid having the version show up there in the spec. Like other things discussed, I think that belongs as part of the discovery design.

mikebrow: note: https://github.com/opencontainers/distribution-spec/blob/master/spec.md#api-version-check

Agree we need long term plan/path for versioning, both for distribution and it's extensions.

ext/README.md
```diff
@@ -0,0 +1,49 @@
...
+* [Prior Art](#prior-art)
+<!-- /toc -->
+
+## Table
+
+_notice_: All new `./ext/ext-$name.md` docs MUST be added to this table.
+
+| `$name`  | Definition | Summary |
+|:--:|:--:|:--:|
+| 0 | [ext-0](./ext-0.md) | Base definition of discovering extensions on registry server |
```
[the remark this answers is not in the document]
vbatts: > s/Base definition of/Extension for/
> 
> suggest removing definition column... and using link`[0](./ext-0.md)`
> 
> Suggest columns for number, title, and summary description

okie

> Suggest using the file name:`0-index-of-distribution-extensions.md`

Fair, but i don't expect the `$name` to be only numbers. Just that 0 is the first either way you sort. And when globbing for _only_ extensions, `ext-*.md` is cleaner than excluding a README.md

ext/README.md
mikebrow: maybe reserve prepended `_` for registry operations (legacy) and prepend `_ext-[0-9]+` for extensions listed here such as `_ext-0`

ext/README.md
vbatts: yes?
I looked at the kube CRD versioning, but didn't care to pull that in just yet.

ext/README.md
josephschorr: I think we should try to limit the number of new top-level paths, as they can easily conflict with paths already exposed by registries. I'd prefer if we kept all extensions under the registry root path, whether that be `/v2` or something else moving forward

spec.md
```diff
@@ -840,6 +841,25 @@ If the image had already been deleted or did not exist, a `404 Not Found` respon
...
 
+### Extensions
+
+Extensions to the distribution-spec's registry API are for the purpose of defining OPTIONAL endpoint behaviour.
+Extensions are only complementary to the distribution-spec only having additive features.
+Client can discover the extensions available, if any.
+
+These extensions are similar concept to the use of "feature flags".
+
+See the [`./ext/` directory](./ext/) for extensions available and their definitions.
```
josephschorr: @vbatts Do you have any thoughts on having an API-driven discovery mechanism for back-compatibility between versions of extensions, or do you fully intend all new versions of extensions to be completely independent?

ext/README.md
vbatts: @dmcgowan are you saying you would prefer `/v2/_ext/$name/...`?
@josephschorr is ^ inline with what you're saying?

keep in mind that `$name`-extension would be something like "signatures" for notary-v2, or "events" for the pub-sub proposal.

spec.md
vbatts: is this to say, clients could stay using the older version of an extension?
If so, no. That is up to the registry maintainer. I think the `$name` of any given extension ought to link to the its markdown doc describing it. If it changes, then the `$name` and doc should change. The registry owner can choose the lifecycle they support here.

ext/README.md
dmcgowan: Yeah, `/v2/_ext/...` works. It's not pretty but we can't override the namespace there. I agree with @josephschorr on limiting the number of top level paths. Making them ugly helps I guess ;)

ext/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Table
+
+_notice_: All new `./ext/ext-$name.md` docs MUST be added to this table.
+
+| `$name` (and definition) | Summary |
+|:--:|:--:|
+| [0](./ext-0.md) | Extensions discovering extensions on registry server |
+|  |  |
+
```
vbatts: "by virtue of being added to this table 'reserves' the path `/v2/_ext/$name/`"

spec.md
josephschorr: Hmm... I have concerns we'll run into the same problems that we're running into with `/v2` in regards to back and forward compatibility. At minimum, I think we should have extensions have a defined way of publishing support for specific things or versions

vbatts:
from a discussion that @dmcgowan and I had today.

While I thought that all the endpoints of any given extension would be sandboxed in its prefix, Derek points out how that could make the ACLs a bit too tricky, without wholly expecting an extension to copy API patterns.

So we walked through both.
We used the notary v2 use-case as our first concrete example.

Notes below (copied from https://hackmd.io/Ljkrt0LORmeVznREeGOWuQ):
```HTTP
GET /v2/_ext/
```

```JSON
{
  "0" : {},
  "sigs": {"doc":"http://github.com/opencontainers/distribution-spec/trees/master/ext/ext-sigs.md"}
}
```

```HTTP
/v2/_ext/sigs/   (reserved for sigs extension)
```

```HTTP
GET /v2/_ext/sigs/<hash>/<digest>/<name>
POST /v2/_ext/sigs/<hash>/<digest>/<name>
DELETE /v2/_ext/sigs/<hash>/<digest>/<id>

GET /v2/<repo name>/sigs/<hash>/<digest>
POST /v2/<repo name>/sigs/<hash>/<digest>
PUT /v2/<repo name>/sigs/<hash>/<digest>/<hash>/<digest>
DELETE /v2/<repo name>/sigs/<hash>/<digest>/<hash>/<digest>
                            |<--signed---->|<--signature-->|       
```

```JSON
{
  "0": {
    "mediaType":"...signature...blob",
    "size": 100,
    "digest": "sha256:deadbeef"
  }
}
```
What ever mediaType of the signature object itself should be parseable to reference the digest of the blob that it is a signature for. (for garbage collection)

josephschorr:
@vbatts The concern I have there is overlap: if I currently have a registry that supports multiple levels of nesting, and I have a repository named `mynamespace/arepo/sigs` (for whatever reason), now I have a conflict on the paths. Does that means I get the repository responses back, or do I get the `sigs` for `mynamespace/arepo` (it might exist on its own, afterall). A similar problem does exist today, but its currently limited in the API, where moving forward this wouldn't be the case.

Perhaps all extensions under the existing API should be prefixed with a character not supported by the current repo names? Like... `GET /v2/<repo name>/_/sigs/<hash>/<digest>`?

dmcgowan:
@josephschorr when the API was originally being designed a similar issue also came up with `tags`, hence the odd `tags/list` endpoint for listing tags. It may have been a limitation of the handlers we were using at the time, but there was always two parts matched at the end to get around the ambiguity (@stevvooe for more context). I kind of like the idea of differentiating with a "_", maybe "_sigs" is enough. As long as we apply the rule consistently for new endpoints.

jzelinskie:
Why not just adopt something like the `.well_known` and let the servers define their own URL prefixes and try and stay as much out of their way as possible? Imposing versioning and all of this other stuff in the URL seems pretty non-ideal.

josephschorr:
@jzelinskie I'm highly concerned of overlap: its *guaranteed* that registries, today, have `/v2/...` reserved for registry use. If we allow extensions to place themselves wherever they please at the top level, we will likely quickly run into conflicts with existing per-implementation endpoints. `.well-known` is a good idea, but it should ideally be discoverable as part of the existing OCI endpoints.

samuelkarp:
For new extensions that are being proposed, but have not yet been accepted into the spec, do we want to set a standard for naming that distinguishes a proposed-extension from a standard-extension?  I'm thinking along the lines of [vendor prefixes for CSS](https://developer.mozilla.org/en-US/docs/Glossary/Vendor_Prefix#CSS_prefixes); this could allow a registry operator to experiment and provide an endpoint for a feature where there isn't yet consensus in the community.

sequix:
What will this extension covers? Everything in the [brainstorm](https://hackmd.io/@cyphar/ociv2-brainstorm#)?

spec.md
```diff
@@ -33,6 +33,7 @@ The goal of this specification is to standardize container image distribution ba
   - [Listing Repositories](#listing-repositories)
```
jdolitsky, line 33: maybe we hold off on mods to spec.md until the reorg is merged: https://github.com/opencontainers/distribution-spec/issues/163

vbatts:
scope of the extensions:

     extensions are meant as a way to extend on the generic basic workflows available in the distribution-spec

A way to have a handshake between client-server for additional features capable. e.g. search, signatures, discoverability. etc

The hope of having a server expose which features it implements vs. a newer version (i.e. `/v2/`) that implements new features, in synchrony.
It's okay to add a batch of OPTIONAL features to the spec, with language around "`if you use this OPTIONAL http endpoint then you MUST use the corresponding ones`".

I'm really wondering whether this is even the correct approach.

If these features are the right and good for distribution-spec (the API), then just put them in the spec and rev the version.


(Having a call with @dmcgowan @jzelinskie @mikebrow @jdolitsky @vbatts)

vbatts:
This basically needs community discussion around "how can i experiment with bringing in new features, or have my own bespoke feature"

spec.md
SteveLasker: would `/v2/_ext/oci-artifacts/v1/{namespace}/repo/manifests/{digest}/links?n=10` be inline with what we're discussing here?
- oci-artifacts would have a links extension.
- it would have a version `v1`
- it would follow the `namespace/repo` path consistency, but be rooted in a named extension `/oci-artifacts`

ext/README.md
sajayantony: 1. What would be the guidance be for extension authors to avoid naming collisions/squatting?
Should we consider  `/v2/_ext/oci-artifacts` or `/v2/_ext/signatures` or `/v2/_ext/oci/signatures`.

2. Also would providing guidance around `/v2/_ext` itself. Is there an expectation of returning a list of supported extensions or are the names something that is discovered out of band and we just 404 on `/v2/_ext/oci/signatures`

SteveLasker:
What was the consensus here?
At the root:
`/v2/_ext/oci-artifacts/v1/{namespace}/repo/manifests/{digest}/links`
or, at the repo
`/v2/{namespace}/repo/manifests/{digest}/_ext/oci-artifacts/v1/links`

Basically, should `_ext/` be a reserved extension anywhere in the uri?

SteveLasker:
@vbatts, @dmcgowan, @justincormack, in today's OCI weekly call, we discussed moving forward with https://github.com/opencontainers/distribution-spec/pull/111/ where we document `v2/_ext/` as a reserved namespace for extensions. The question was then: should we use numbered items, and let people fight for reserving a number while they develop and hopefully ship their extension. Or, named elements that have no order. Just uniqueness.

vbatts:
now that 1.0 is tagged, and a number of conversations that touched on this proposal have progressed, I intend to dust this off and find the path forward.

SteveLasker:
Thanks @vbatts, here's a prototypical example for the reference types: https://github.com/opencontainers/artifacts/pull/29
A direct [link to the referrers API](https://github.com/opencontainers/artifacts/blob/d909db51eb08e9da278bba99bddc1c878f1a0ed3/manifest-referrers-api.md)

stevvooe:
I like the idea here but getting the paths right can be difficult. The problem we solved with `tags/list` was that you needed something to anchor the path grammar. Something like `/tags/` was just as effective, since we have a terminating slash (there was a proposal for this at one time) but it is subtle. This is probably overly conservative but prevents future incompatibilities while still namespacing operations beneath a repository. Without getting into the finer points of why putting user input into a path is a bad idea, it works well enough. 👏 

In general, I always saw `/v2/_<thing>`, such as `/v2/_catalog` as extensions to the API. Early notary proposals had something like `/v2/_notary`. In reality, there are two main things we need to cover:

1. Registry-level extensions.
2. Repository-level extensions.

That underscore is ugly, indeed, but isn't supported in the repository name grammar so works well as an escape. We can safely introduce the following two path components to the specification and get away with it:

1. `/v2/_<ns>/<ext>/<component>[?<key>=<value>&...]` would provide access to a registry-level extension.
2. `/v2/<name>/_<ns>/<ext>/<component>[?<key>=<value>&...]` would provide access to repository-level extensions.

This satisfy requirements for namespacing, for the most part. The path syntax for the namespace is limited to keep things simple. We restrict extensions to query parameters. __Paths are a nightmare for user input and query parameters reduce these ambiguities.__ I'd favor the simple approach of keeping it to the above anchor points, meaning that we have registry and repo level extensions.

Let's roll through that exercise a bit. From the proposed artifacts extension:

`/oci-artifacts/v1-rc1/{repository}/manifests/{digest}/referrers?n=10`

This would become something like this:

`/v2/<name>/_oci/artifacts/referrers?digest={digest}&n=10`

Notice we can elide the version number because we use the mediatypes to version content in oci. We might want to define some conventions for pagination and assumed object references to keep things consistent.

I think one thing that needs to be worked out is whether extensions will be a free for all or more like an a la carte set of APIs that registries can choose to implement or not. I really favor the a la carte model, as it ensures that registries can choose which features make sense for their use case, but will ensure compatibility between extensions. I do understand the desire to have an open extension model, however.

It might a good idea to take these examples through a few more use cases to see how they work out. I want to make sure we have some strong conventions for additions, follow the model of versioned data types but still allow for objects that exist outside the artifact/image's CAS lineage.

sajayantony:
+1

The [referrers API](https://github.com/oras-project/artifacts-spec/blob/main/manifest-referrers-api.md) is currently rooted under  `{host}/oras/artifacts/v1/`. Implementers including our own internal deployments run distribution and expect their path to be `{host}/v2/` for different ACLS or ingress rules. 

Providing an ability for extension `/v2/<name>/_oci/artifacts/referrers?digest={digest}&n=10` will enable extension to be added without needing to open up new paths on the ingress controllers. 

My hope here is that if we get agreement on this issue then there is a possibility of queuing up support into the reference implementation like CNCF/distribution. 

I would image that `/v2/<name>/_oci/` would be reserved for OCI specific extensions and extension authors can follow the same pattern. like `/v2/<name>/_oras/.

sudo-bmitch (changes requested):
Fantastic update. Should we clarify that individual extensions are optional for any registry to implement?

ext/ext.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Reference Explanation
+
+The base extension may be queried through a `GET` to discover available
+extensions.
+
+Registry level extensions may be discovered with a standard GET as follows.
+
+```HTTP
+    GET /v2/_ext/
```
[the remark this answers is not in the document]
mnltejaswini: ```/v2/_oci/ext``` , is oci considered as the default namespace for all registry operations?

ext/ext.md
```diff
@@ -0,0 +1,56 @@
...
+endpoints as shown below.
+
+```HTTP
+200 OK
+Content-Length: <length>
+Content-Type: application/json
+
+{
+    "extensions: [
+        "_<ns>/<ext>/<component>",
```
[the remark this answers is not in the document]
sudo-bmitch: Sorry, missed a slash in my comment, fixed. This is for the response to the API request. If we return an array of strings, it can't be easily extended later if we want to include other metadata about extensions.

ext/ext.md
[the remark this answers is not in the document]
sudo-bmitch: I see OCI as the goal for things added that aren't vendor specific. They should be used by extensions that have been standardized/approved by OCI. If a specific registry implementation adds their own extensions, they should be doing so under their own namespace.

ext/README.md
```diff
@@ -0,0 +1,78 @@
...
+---
+
+# Extensions API for Distribution
+
+The basis of the Extension API is described in a document which should
+be emulated for all extensions.
+
+## Table
+
+_notice_: All new `./ext/ext-$name.md` docs MUST be added to this table.
```
stevvooe, line 13: Was our goal to have a centralized registry of extensions? Or, is this for standardized extensions?

ext/ext.md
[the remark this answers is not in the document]
stevvooe: I think if we follow these conventions, it would be something like `/v2/_ext/discover`, since the spec sits on the `_ext` namespace to surface extensions. This would also allow the spec to add other extension operations in the future.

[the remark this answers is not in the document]
SteveLasker: For the vendor-specific question:
All registries build extensions, to fill the gaps in the specs. Since the cloud-hosted registries run on their domains, apis outside of `/v2/` hasn't been a problem.
What came up with the ORAS Referrers API is how this would get implemented by harbor, cncf distribution, jfrog, where they can run their registries in a users k8s cluster. Having APIs be out of the `/v2` path provides routing challenges. 

The premise around the extensions is to enable optional OCI capabilities, but also enable registry-specific capabilities to have a well-defined location to run within. If harbor, gcr, acr, ecr, jrog, zot are implemented on these paths, they just need names that keep them unique. For instance, Alibaba Container Registry is also acr :), so azure/acr might be a path acr might use.
The point is I think we do need a well-defined way for optional extensions, whether they're vendor-specific extensions or community extensions like a scan result indexing tool that @hectorj2f mentions. k8s has an extension model, how can we apply one to the distribution spec?

ext/README.md
sajayantony: I was not expecting this doc to be the clearing or mechanism for registration of extensions but community could submit extensions to this repo. For now I was only considering the discovery extension to be in this location. 
./ext/ext-oci.md which would call out the definition for `/v2/_ext/oci/extensions/list` and `/v2/<repo>/_ext/oci/extensions/list` or similar discovery APIs.

sudo-bmitch (dismissed):

vbatts:
I've updated the [OP comment](https://github.com/opencontainers/distribution-spec/pull/111#issue-386853850) to outline the goals of this PR.

My very original thought was to have a framework that is as flexible as the concept of HTTP Headers: there are some expected, arbitrary optional ones, but some have grown to be established norms.

After staring at maintaining the goals here, it's not clear that the addition/experimentation of any given extension is on the whole any different than folks just _adding_ OPTIONAL routes to their registry, and then proposing them to the core of OCI distribution-spec once they're tested and refined.

SteveLasker:
Thanks @vbatts for the update. Capturing the permissions challenges with registry wide, a namespace, or specific to a repo helps define the requirements for the extensions.

For example:
- docker hub, github and gcr use a single domain, with a root namespace to delineate "registry ownership" `docker.io/org`, `gcr.io/project/repo`
- acr, ecr and others use a domain to differentiate a registry `[myregistry].azurecr.io`, `[aws_account_id].dkr.ecr.us-west-2.amazonaws.com/`
- An auth-token may only have access to a specific repo, `registry.io/org/team/repo`, so making calls to `registry.io/_foo` raises security challenges.

vbatts:
from @stevvooe on slack

> there is an extension endpoint hanging off of /v2 , a registry extension, and an extension endpoint hanging off of /v2/<name> , a repository extension. They each have discovery extensions. If the endpoint for the registry is called, no repository name is included. if the endpoint for the repository is called, then the repository name is included. This allows one to calculate the full extension url from the json.

vbatts:
https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints

vbatts:
@stevvooe right on. That covers ACLs.

This implies that if any extension that deals in the `/v2/<name>/_...` space, but may work with blobs, digests and/or manifests, then the extension may recreate/emulate existing endpoints in the core API. Which is a bit sloppy to me, but is "flexible" for folks.

stevvooe:
I wanted to summarize more the discussion we had on slack regarding extensions:

1. **Paths**: Extension paths are a slash-separated triple, consisting of a `namespace`, `component` and `operation`, in the format of `_<namespace>/<component>/<operation>`.
  a. **Namespace**: A `namespace` corresponds to a single extension or set of extensions.
  b. **Format**: Path components must follow the same restrictions as repository name components.
  c. **Definition**: We don't define the meaning of `component` and `operation`. Extension providers may organize them as they see fit but are recommend to organize these in a way that allows their extension to change over time.
  d. **Example**: For the proposed discovery endpoint, the triple would be `_oci/ext/discover`. We can say that "the OCI extension provides a discover endpoint for extensions."
2. **Path Mounting in API**: There are two places that an extension may appear in the path hierarchy:
  a. **Registry-level Extension**: A registry-level extension is mounted at the root of the `/v2/` path. The full path format would be `/v2/_<namespace>/<component>/<operation>`
  b. **Repository-level Extension**: A repository-level extension is mount at `<name>`, as a sibling of `blobs` and `manifests` of the `/v2/<name>` path. The full path format would be `/v2/<name>/_<namespace>/<component>/<operation>`
3. **Discovery**: The discovery extension provides discovery of extension endpoints, identified by the triple `_oci/ext/discover`. It will also act as an example of how to implement an extension in an idiomatic way.
  a. **Registry-level**: The `/v2/_oci/ext/discover` will provide a list of extensions available at the registry level. It will respond to GET requests with the following format:
          ```{
        "extensions": [
          {
            "path": "_<namespace>/<component>/<operation>"
          }, ...
        ]
      }```
  b. Paths for the extension can be assembled by joining `/v2/` with the extension `path`, such that the complete path would be in the format `/v2/_<namespace>/<component>/<operation>` for each extension entry.
  c. **Registry-level**: The `/v2/<name>/_oci/ext/discover` will provide a list of extensions available at the repository level. It will respond to GET requests with the following format:
          ```{
        "repository": "alpine",
        "extensions": [
          {
            "path": "_<namespace>/<component>/<operation>"
          }, ...
        ]
      }```
  d. Paths for the extension can be assembled by joining `/v2/<name>` with the `path`, where `name` is specified in the `repository` field, such that the complete path would be in the format `/v2/<name>/_<namespace>/<component>/<operation>` for each extension entry.

I know this is a bit dense but hopefully I've covered all of the points.

stevvooe:
For the above, I prefer the word `<module>` over `<operation>`. Please vote on this comment if we prefer that change.

stevvooe:
Summarizing requirements on registering extensions:
- Add language to have folks register their namespaces to ensure they don't overlap:
```"_notice_: All new `./ext/ext-$ns-$name.md` docs MUST/SHOULD be added to this table."```
- Must include link to documentation, source repository, tests, sensible naming
- Reservation for `_oci` and former extensions (ie `_catalog`)

File: extensions/_oci.md [hunk 11]
```diff
@@ -32,21 +38,25 @@ Content-Type: application/json
...
+  which MAY be constructed by replacing
+  the base discovery extension path,`_ext`.
 
-    This REQUIRED property contains a list of supported extension endpoints.
+  - **`name`** *string*, REQUIRED
 
-    The [relative URI]((https://tools.ietf.org/doc/html/rfc1808))
-    which MAY be constructed by replacing
-    the base discovery extension path,`_ext`.
+    XXX
```
stevvooe, line 59: Looks like we have `path` below. Need to fill that in? ;)

specs-go/v1/ext.go
```diff
@@ -0,0 +1,25 @@
...
+
+package v1
+
+// ExtensionList
+type ExtensionList struct {
+	Extensions []Extension `json:"extensions"`
+}
+
+type Extension struct {
+	Name string
```
stevvooe, line 23: Do these need json tags?

extensions/_oci.md
```diff
@@ -32,21 +38,25 @@ Content-Type: application/json
...
 ### *Extensions* Property Descriptions
 
-- **`extensions`** *array of extension objects with properties like `name`*
+- **`extensions`** *array of extension objects*, REQUIRED
+
+  This property contains a list of supported extension endpoints.
+
+  The [relative URI]((https://tools.ietf.org/doc/html/rfc1808))
+  which MAY be constructed by replacing
+  the base discovery extension path,`_ext`.
```
stevvooe, line 55: While this extension is mounted under `_ext`, I don't think URI is relative `_ext`. Or maybe I'm missing something?

extensions/_oci.md
```diff
@@ -32,21 +38,25 @@ Content-Type: application/json
 {
     "extensions: [
         {
-            "name": "_<ns>/<ext>/<component>"
+            "name": "_<ns>/<ext>/<operation>"
```
stevvooe, line 41: Based on the above, I think this is `_<namespace>/<component>/<module>`.

extensions/README.md
```diff
@@ -5,25 +5,31 @@ breaks: false
...
+Extension names MUST be unique.
+Extensions recorded in this distribution-spec are considered canonical definitions.
+
+Extensions are specified by extension namespace name (`$ext`) aligning with the project, followed by the `component`, and last by by the `module`.
+This constitutes the URI segments of the extension endpoint.
+Additional options may be passed as parameters to the endpoint.
 
 ```http
-_<ns>/<ext>/<component>[?<key>=<value>&...]
+_<ext>/<component>/<module>[?<key>=<value>&...]
```
stevvooe, line 32: Are we going with `_<namespace>/<component>/<module>`?

stevvooe:
```suggestion
_<namespace>/<component>/<module>[?<key>=<value>&...]
```

jdolitsky (dismissed):

extensions/_oci.md
stevvooe:
```suggestion
"name": "_<extension>/<component>/<module>"
```

extensions/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Registry Level Extensions
+
+Registry level extensions are nested under `/v2`.
+
+```http
+GET /v2/_<ns>/<ext>/<component>[?<key>=<value>&...]
+```
+
+For example a search extension may be of the form `/v2/_oci/catalog/search?pattern=foo?n=10`
```
stevvooe, line 45:
```suggestion
For example a search extension may be of the form `/v2/_oci/catalog/search?pattern=foo&n=10`
```

extensions/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Repository Level Extensions
+
+Repository level extensions follow the same naming rules the endpoints
+can be access under a repository.
+
+```http
+GET /v2/<name>/_<ns>/<ext>/<component>[?<key>=<value>&...]
```
stevvooe, line 53:
```suggestion
GET /v2/<name>/_<extension>/<component>/<module>[?<key>=<value>&...]
```

extensions/README.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+For versioning, extensions SHOULD use versioned `mediaType`.
+
+### Registry Level Extensions
+
+Registry level extensions are nested under `/v2`.
+
+```http
+GET /v2/_<ns>/<ext>/<component>[?<key>=<value>&...]
```
stevvooe, line 42:
```suggestion
GET /v2/_<extension>/<component>/<module>[?<key>=<value>&...]
```

extensions/README.md
```diff
@@ -0,0 +1,93 @@
+---
+tags: oci,distribution
+breaks: false
+---
+
+# Extensions API for Distribution
+
+Extensions, in general, to the distribution-spec API are not denied by the distribution-spec.
```
mikebrow, line 8:
```suggestion
Extensions, in general, to the distribution-spec API are allowed by the distribution-spec, with certain reservations disclosed herein.
```

extensions/README.md
```diff
@@ -0,0 +1,93 @@
...
+This document outlines path and safe place for registry implementations and clients to iterate on new APIs, establish them, and propose them to OCI for canonical reference.
```
mikebrow, line 9:
```suggestion
This document outlines a path for registry implementations and clients to iterate on new APIs, establish them, and propose them to OCI for canonical reference.
```

extensions/_oci.md
```diff
@@ -0,0 +1,68 @@
...
+The base extension may be queried through a `GET` to discover available
+extensions.
+
+Registry level extensions may be discovered with a standard GET as follows.
+
+```HTTP
+    GET /v2/_oci/ext/discover
+```
+
+```
```
sajayantony, line 18: I think we can remove this.

extensions/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+The basis of the Extension API (`_oci`) is described in a document which should be emulated for all extensions.
```
mikebrow, line 11:
```suggestion
The basis of the Extension API (`_oci`), as described herein,  should be emulated for all extensions.
```

extensions/README.md
```diff
@@ -0,0 +1,93 @@
...
+```http
+GET /v2/_<ns>/<ext>/<component>[?<key>=<value>&...]
+```
+
+For example a search extension may be of the form `/v2/_oci/catalog/search?pattern=foo?n=10`
+
+### Repository Level Extensions
+
+Repository level extensions follow the same naming rules the endpoints
+can be access under a repository.
```
mikebrow, line 50:
```suggestion
can be accessed under a repository.
```

extensions/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+```http
+GET /v2/_<ns>/<ext>/<component>[?<key>=<value>&...]
+```
+
+For example a search extension may be of the form `/v2/_oci/catalog/search?pattern=foo?n=10`
+
+### Repository Level Extensions
+
+Repository level extensions follow the same naming rules the endpoints
```
mikebrow, line 49:
```suggestion
Repository level extensions follow the same naming rules. The endpoints
```

extensions/README.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+TODO
+```
+GET /v2/vbatts/myapp/_oci/referrers/link_to?
+```
+
+### Reserved Namespaces
+
+As of current, ```_oci``` and ```_ext``` are considered as reserved namespaces and cannot be used by other extensions.
```
mikebrow, line 63:
```suggestion
```_oci``` and ```_ext``` are reserved namespaces and cannot be used by other extensions.
```

extensions/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Filename
+
+Extension definitions SHOULD be placed under `./ext/` in the specification repository. Extension files
```
mikebrow, line 67:
```suggestion
Extension definitions MUST be placed under `./ext/` in the specification repository. Extension files
```

extensions/README.md
```diff
@@ -0,0 +1,93 @@
...
+SHOULD follow the `ext-$ns-$name.md`. Refer [ext.md](./ext.md) for more details.
```
mikebrow, line 68:
```suggestion
MUST follow the `ext-$ns-$name.md`. Refer [ext.md](./ext.md) for more details.
```

extensions/README.md
```diff
@@ -0,0 +1,93 @@
...
+The basis of the Extension API (`_oci`) is described in a document which should be emulated for all extensions.
+
+## Table
+
+_notice_: All new `/extensions/_$ext.md` docs MUST be added to this table.
+
+| Extension (`$ext`)       | Summary                                              |
+|:------------------------:|:----------------------------------------------------:|
+| [`_oci`](./_oci.md)      | Extensions discovering extensions on registry server |
+| `_catalog`               | Resevered prior extension                            |
```
samuelkarp, line 20: nit: typo

```suggestion
| `_catalog`               | Reserved prior extension                            |
```

extensions/_oci.md
```diff
@@ -0,0 +1,73 @@
...
+
+  This property contains a list of supported extension endpoints.
+
+  The [relative URI]((https://tools.ietf.org/doc/html/rfc1808))
+  which MAY be constructed by replacing
+  the base discovery extension path,`_oci`.
+
+  - **`name`** *string*, REQUIRED
+
+    XXX
```
sajayantony, line 60:
```suggestion
```

extensions/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+Repository-level extensions follow the same naming rules as the [registry-level endpoints](#registry-level-extensions).
+
+```http
+GET /v2/<name>/_<extension>/<component>/<module>[?<key>=<value>&...]
+```
+
+### Reserved Namespaces
+
+As of current, `_oci` and `_catalog` are considered as reserved namespaces and cannot be used by other extensions.
```
rchincha, line 57: ... or as part of repository names.

vbatts: oh for sure. In the updated push I include some string formatting regular expression. With the `_` prefix on the extension name, that differentiates all extensions, from any `<name>` of a manifest reference.

vbatts:
updated PTAL

stevvooe (approved):

sajayantony (approved):

extensions/_oci.md
```diff
@@ -0,0 +1,79 @@
...
+
+  - **`url`** *string*, REQUIRED
+
+    URL link to the documentation defining this extension and its endpoints.
+
+  - **`description`** *string*, OPTIONAL
+
+    Human readable description of this endpoint.
+
+  - **`endpoints`** *array of strings*, REQUIRED
```
sudo-bmitch, line 65: Are "endpoints" a reference to "components"?

vbatts: ah ok. we need examples

vbatts:
@opencontainers/distribution-spec-maintainers PTAL

extensions/_oci.md
sudo-bmitch: For anyone that wasn't on the call today, endpoints refers to `_<extension>/<component>/<module>`. Not sure if we want to include `/v2/` and `/v2/<repository>` in there.

mikebrow (approved):
LGTM for initial push

vbatts:
:clap:

sartoripeter044-jpg (approved):

PR #125, opened 2020-04-01
WIP: spec rewrite

jdolitsky:


vbatts:
During the call today: 
Generally everyone is :+1: on the layout. Concern is the drastic nature, so we don't have an impossible review, especially if there is _new_ content added during, so for now lets make a tracker for this, and break this into incremental chunks (just reorganized).

vbatts:
https://github.com/opencontainers/distribution-spec/issues/126 to track

jdolitsky:
Will close and resubmit in smaller chunks

PR #129, opened 2020-04-06
Modify manifest config to conform to image-spec

### LINKED ISSUE #115: Conformance test's config does not conform to spec
According to the image spec, it appears the `rootfs` property is *required* in the config: https://github.com/opencontainers/image-spec/blob/master/config.md:

```rootfs object, REQUIRED```

However, it appears the conformance test suite is sending an empty configuration block:
https://github.com/opencontainers/distribution-spec/blob/4c5b810d62dfbe50310356019e598827a56312ed/conformance/setup.go#L92

Does the conformance suite need to be updated or is the intention for the `rootfs` property to be optional and assumed empty if not present?

pmengelbert:
Resolves #115

### DIFF
--- a/conformance/setup.go
+++ b/conformance/setup.go
@@ -119,7 +119,16 @@ func init() {
 
 	client.SetLogger(logger)
 
-	configContent = []byte("{}\n")
+	configContent = []byte(`
+{
+    "architecture": "amd64",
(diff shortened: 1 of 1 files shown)


vbatts:
LGTM

PR #130, opened 2020-04-07
Enable user to hide skipped workflows in report

### LINKED ISSUE #128: conformance: remove skipped workflow categories from report
If a registry doesn't support a workflow, that workflow should be removed from the report to reduce clutter and increase the proportion of green results.

pmengelbert:
Skipped workflows can be hidden from the report by setting
`OCI_HIDE_SKIPPED_WORKFLOWS=1` in the environment.

Also improves whitespace in HTML template string.

Resolves #128

### DIFF
--- a/conformance/01_pull_test.go
+++ b/conformance/01_pull_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 var test01Pull = func() {
-	g.Context("Pull", func() {
+	g.Context(titlePull, func() {
 
 		var tag string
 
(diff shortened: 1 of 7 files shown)


pmengelbert:
> LGTM
> nit... a lot of whitespace changes :-)

@mikebrow
The whitespace reorganization was intentional, it was getting out of hand in the HTML template string. I amended the commit message to reflect the changes

jdolitsky:
should be good to merge, just cosmetic changes here

vbatts:
LGTM

PR #133, opened 2020-04-09
Upload a layer to be included in the manifest

### LINKED ISSUE #118: conformance: do not push empty layers
cc @jonjohnsonjr @samuelkarp

pmengelbert:
There is a base64 enoded string which, when decoded, contains a tarball.
This is uploaded as a blob and then included in the `"layers": []`
section of the manifest. Some registries were rejecting the manifest due
to an empty layers section, so this fix should resolve that issue.

Resolves #118

### DIFF
--- a/conformance/02_push_test.go
+++ b/conformance/02_push_test.go
@@ -29,7 +29,7 @@ var test02Push = func() {
 
 				req = client.NewRequest(reggie.PATCH, resp.GetRelativeLocation()).
 					SetHeader("Content-Type", "application/octet-stream").
-					SetBody(blobA)
+					SetBody(testBlobA)
 				resp, err = client.Do(req)
 				Expect(err).To(BeNil())
 				Expect(resp.StatusCode()).To(Equal(http.StatusAccepted))
@@ -39,9 +39,9 @@ var test02Push = func() {
 			g.Specify("PUT request to session URL with digest should yield 201 response", func() {
 				SkipIfDisabled(push)
 				req := client.NewRequest(reggie.PUT, lastResponse.GetRelativeLocation()).
-					SetQueryParam("digest", blobADigest).
+					SetQueryParam("digest", testBlobADigest).
 					SetHeader("Content-Type", "application/octet-stream").
-					SetHeader("Content-Length", blobALength)
+					SetHeader("Content-Length", testBlobALength)
 				resp, err := client.Do(req)
 				Expect(err).To(BeNil())
 				Expect(resp.StatusCode()).To(Equal(http.StatusCreated))
@@ -63,10 +63,10 @@ var test02Push = func() {
 			g.Specify("POST request with digest and blob should yield a 201", func() {
 				SkipIfDisabled(push)
 				req := client.NewRequest(reggie.POST, "/v2/<name>/blobs/uploads/").
-					SetHeader("Content-Length", configContentLength).
+					SetHeader("Content-Length", configBlobContentLength).
 					SetHeader("Content-Type", "application/octet-stream").
-					SetQueryParam("digest", blobDigest).
-					SetBody(configContent)
+					SetQueryParam("digest", configBlobDigest).
+					SetBody(configBlobContent)
 				resp, err := client.Do(req)
 				Expect(err).To(BeNil())
 				location := resp.Header().Get("Location")
@@ -76,7 +76,7 @@ var test02Push = func() {
 
 			g.Specify("GET request to blob URL from prior request should yield 200", func() {
 				SkipIfDisabled(push)
-				req := client.NewRequest(reggie.GET, "/v2/<name>/blobs/<digest>", reggie.WithDigest(blobDigest))
+				req := client.NewRequest(reggie.GET, "/v2/<name>/blobs/<digest>", reggie.WithDigest(configBlobDigest))
 				resp, err := client.Do(req)
 				Expect(err).To(BeNil())
 				Expect(resp.StatusCode()).To(Equal(http.StatusOK))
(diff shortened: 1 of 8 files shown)


pmengelbert:
@jonjohnsonjr

Will this be sufficient to get GCR to accept the manifest upload?

pmengelbert:
@jonjohnsonjr @mikebrow 
This PR enables GCR to pass the pull workflow

conformance/setup.go
```diff
@@ -129,36 +145,57 @@ func init() {
...
+	if err != nil {
+		log.Fatal(err)
 	}
 
-	manifestContent = []byte(fmt.Sprintf(
-		"{ \"mediaType\": \"application/vnd.oci.image.manifest.v1+json\", \"config\":  { \"digest\": \"%s\", "+
-			"\"mediaType\": \"application/vnd.oci.image.config.v1+json\","+" \"size\": %s }, \"layers\": [], "+
-			"\"schemaVersion\": 2 }",
-		blobDigest, configContentLength))
+	layerBlobDigest = godigest.FromBytes(layerBlobData).String()
```
mikebrow, line 159: If we delete all tests with issues... we won't have a way to track them :-) IMO we should lean to adding additional tests and have a config that enables skipping over the tests that are not supported on some registries due to "disagreement" on container standards till we can reach agreement. I believe the empty layer test to be a good test. Thoughts?

pmengelbert: @mikebrow 
Happy to implement that way. I can return the manifest config to the way it was (empty layer) and only add the layer when a certain environment variable is set.

mikebrow: sure or just have two tests one empty and one not and use a ginkgo regex to skip empty for the GCR registry

pmengelbert: I went with the former, since when I got your response the solution was already underway :-D.

mikebrow: by skip empty for GCR I meant let them configure it to skip if that's what they want but we can start with an initial config that maps out what we see passing and what needs to be skipped..

jdolitsky:
@mikebrow couple thoughts-

The spec doesn't explicitly state the acceptable/required number of layers for upload. That being said, a single layer appears to meet the requirements of all registries, and seems like a reasonable expectation.

I'm in favor of doing only one test - either as is (0 layers) or this change (1 layer). I dont find it useful to add a flag here to disable a no-layer test as it complicates the test matrix even further. Now instead of 4 categories, one of the categories (push) would be broken down further.

I suggest we remove this as a configurable option, and modify the spec language the say something along the lines of "registries MUST accept one (1) layer, but MAY provide zero (0) layers or multiple layers". 

If the spec will end up reading that accepting 0 layers is required, then we should just do both tests and GCR considered non-conformant for push.

mikebrow:
> @mikebrow couple thoughts-
> 
> The spec doesn't explicitly state the acceptable/required number of layers for upload. That being said, a single layer appears to meet the requirements of all registries, and seems like a reasonable expectation.
> 
> I'm in favor of doing only one test - either as is (0 layers) or this change (1 layer). I dont find it useful to add a flag here to disable a no-layer test as it complicates the test matrix even further. Now instead of 4 categories, one of the categories (push) would be broken down further.
> 
> I suggest we remove this as a configurable option, and modify the spec language the say something along the lines of "registries MUST accept one (1) layer, but MAY provide zero (0) layers or multiple layers".
> 
> If the spec will end up reading that accepting 0 layers is required, then we should just do both tests and GCR considered non-conformant for push.

4x1 matrix? I just assumed/imagined a test tree with branches for the cases where a MUST is not specified than a matrix with 4 big switches and only test paths that are the exact same for all registries.

I don't see much use for the zero layers containers test .. One could argue there can only be one "scratch" I guess.  My point was why delete a test that successfully found an issue/differences between the registries.  Flag it ... comment it.. todo it.. Just not a fan of deleting tests unless there is clear documentation / specification that the test is wrong. More a general thought process around testing than anything else.

Cheers, Mike

sajayantony:
Since distribution can be used for more than images, I would like to add that we **do** use zero layers to store config only artifacts. 
Layers are an overhead when you want a single JSON config blob that can be used as configuration or metadata in the system.

mikebrow:
> Since distribution can be used for more than images, I would like to add that we **do** use zero layers to store config only artifacts.
> Layers are an overhead when you want a single JSON config blob that can be used as configuration or metadata in the system.

Interesting.. thx.

conformance/setup.go
```diff
@@ -19,6 +19,30 @@ type (
 		Name string   `json:"name"`
 		Tags []string `json:"tags"`
 	}
+
+	Manifest struct {
```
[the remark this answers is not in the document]
pmengelbert: Sure -- will include this in an amendment later today

pmengelbert:
@mikebrow 
I just pushed an update that accomplishes 2 things:
1. Use the structs from image-spec
2. By default, run two tests: one with a layer, and one without. If `OCI_SKIP_EMPTY_LAYER_PUSH_TEST=1` is set in the environment, it will skip the empty layer test.

conformance/02_push_test.go
```diff
@@ -9,6 +9,10 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+const (
+	emptyLayerTestTag = "emptylayer"
+)
```
jdolitsky, line 14: Can we put this with other constants in setup.gp?

pmengelbert: Sure

jdolitsky:
This should be good to merge. By default running both 0 layer and 1 layer, with the ability to skip the 0 layer test via `OCI_SKIP_EMPTY_LAYER_PUSH_TEST=1`

dmcgowan:
This one needs a rebase after merging the other pull PR

pmengelbert:
@dmcgowan Just rebased it, should be good to go now

pmengelbert:
@dmcgowan Can we get you to approve again? I think the rebase undid your approval. Should be good to merge now!

PR #144, opened 2020-04-24
Disable cookie handling by default

wy65701436:
Set the cookie jar to nil is to disable save cookies in api contacting.

Signed-off-by: wang yan <wangyan@vmware.com>

### DIFF
--- a/conformance/setup.go
+++ b/conformance/setup.go
@@ -152,6 +152,7 @@ func init() {
 	}
 
 	client.SetLogger(logger)
+	client.SetCookieJar(nil)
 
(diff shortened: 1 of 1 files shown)


dmcgowan:
Maybe this should be the default and if there is a registry in the future which requires cookies to work, then enable it specifically. Any sort of user/session tracking is not mentioned by the spec, so disabling it by default makes sense to me.

wy65701436:
make sense, code is updated per comments.

vbatts:
LGTM

PR #153, opened 2020-05-21
update docker run conformance test command

wy65701436:
The command in the README is not executable, add the missing backslash

Signed-off-by: wang yan <wangyan@vmware.com>


jdolitsky:
LGTM

PR #156, opened 2020-06-05
conformance test: tear down manifests before blobs

majewsky:
As explained in the added comments, "some registries" (i.e. my own implementation, Keppel <https://github.com/sapcc/keppel>) forbid deleting blobs that are still referenced by manifests. To do a proper cleanup in this case, the manifest needs to be deleted first.

The diff looks a bit weird, but I really just swapped the order of some paragraphs, pulling the manifest deletion to the top.


majewsky:
> This PR should be modified to switch the order based on the presence of some env var

Done.

majewsky:
Documentation added.

conformance/01_pull_test.go
```diff
@@ -152,6 +153,21 @@ var test01Pull = func() {
 		})
 
 		g.Context("Teardown", func() {
+			deleteManifestFirst, _ := strconv.ParseBool(os.Getenv(envVarDeleteManifestBeforeBlobs))
```
jdolitsky, line 156: looking through this one more time - could we move this env var check to run only once at the bottom of setup.go? Check something like `runContentDiscoverySetup` , it is initialized as a global then set based on env

majewsky: done

jdolitsky:
@majewsky looks good. Last thing is the DCO check failed on your last commit. Can you amend the commit to contain signoff then we can merge?

jdolitsky:
@majewsky - thanks for you patience, looks like the final commit had a different email in signed-off-by which might be causing an error: https://pullapprove.com/opencontainers/distribution-spec/pull-request/156/requirements/

jzelinskie:
LGTM

PR #160, opened 2020-06-10
mini update 3: fill in overview section

jdolitsky:
Fill in the overview section, including new introduction
referring to "content" (vs. just container images).

Move historical context section underneath overview.

---

This PR is built on top of #159, so the diff is misleading. Please see the following for a true diff:  https://github.com/bloodorangeio/distribution-spec/compare/mini-update-2...bloodorangeio:mini-update-3

PR #161, opened 2020-06-10
mini update 4: fill in definitions section

jdolitsky:
Fill in the definitions section, introducing several
terms and defintions which will serve as reference for
the rest of the document.

Also moved the "notational conventions" section into
this section, removing any language that seemed
unnecessary or related to compliance (which will be
covered in the conformance section).

---

This PR is built on top of #160, so the diff is misleading. Please see the following for a true diff:  https://github.com/bloodorangeio/distribution-spec/compare/mini-update-3...bloodorangeio:mini-update-4


spec.md
```diff
@@ -5,76 +5,63 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
+
+- **Registry**: a HTTP service which implements this spec
+- **Client**: a tool that communicates with registries over HTTP
+- **Push**: the act of uploading content to a registry
+- **Pull**: the act of downloading content from a registry
+- **Artifact**: a single piece of content, made up of a manifest and one or more layers
+- **Manifest**: a JSON document which defines an artifact
+- **Layer**: a single part of all the parts which comprise an artifact
+- **Config**: a special layer defined at the top of a manifest containing artifact metadata
+- **Blob**: a single binary content stored in a registry
```
rchincha, line 58: Do we want to distinguish between "Layer" and "Blob" here?

spec.md
```diff
@@ -5,76 +5,63 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
+- **Digest**: a unique blob identifier
+- **Content**: a general term for content that can be downloaded from a registry (manifest or blob)
```
rchincha, line 60: ... and addressable by its hash?

spec.md
jdolitsky: @rchincha how would you best describe the difference?

spec.md
```diff
@@ -5,76 +5,63 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
+- **Registry**: a HTTP service which implements this spec
+- **Client**: a tool that communicates with registries over HTTP
+- **Push**: the act of uploading content to a registry
+- **Pull**: the act of downloading content from a registry
+- **Artifact**: a single piece of content, made up of a manifest and one or more layers
+- **Manifest**: a JSON document which defines an artifact
+- **Layer**: a single part of all the parts which comprise an artifact
+- **Config**: a special layer defined at the top of a manifest containing artifact metadata
+- **Blob**: a single binary content stored in a registry
+- **Digest**: a unique blob identifier
```
adamwg, line 59: I would call out here that a digest is calculated from the content of a blob, as opposed to being a generic identifier. That's described in more detail later in the spec, but it's important enough to mention early on.

spec.md
jdolitsky: added

spec.md
jdolitsky: added

spec.md
```diff
@@ -41,14 +41,27 @@ For relevant details and a history leading up to this specification, please see
...
+- **Content**: a general term for content that can be downloaded from a registry (manifest or blob)
 
-An implementation is not compliant if it fails to satisfy one or more of the MUST, MUST NOT, REQUIRED, SHALL, or SHALL NOT requirements for the protocols it implements.
-An implementation is compliant if it satisfies all the MUST, MUST NOT, REQUIRED, SHALL, and SHALL NOT requirements for the protocols it implements.
+### Document Language
```
mikebrow: pr needs rebase

spec.md
mikebrow: the definitions appear to be circular, use words not defined.. For example:

layer is a part in parts (part not defined) which comprises an artifact..

artifact is a piece of content (content is content that is addressable and can be downloaded choose one: manifest or blob) made up of a manifest and one or more layers (one or more undefined parts)

blob is binary content (content is content -- either manifest or blob)

spec.md
```diff
@@ -5,76 +5,56 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
+
+Several terms are used frequently in this document and warrant basic definitions:
+
+- **Registry**: a HTTP service which implements this spec
+- **Client**: a tool that communicates with registries over HTTP
+- **Push**: the act of uploading content to a registry
+- **Pull**: the act of downloading content from a registry
+- **Artifact**: a single piece of content, made up of a manifest and one or more layers
+- **Manifest**: a JSON document which defines an artifact
+- **Layer**: a single part of all the parts which comprise an artifact
```
mikebrow, line 53: Layer explanation from image spec: https://github.com/opencontainers/image-spec/blob/master/config.md#oci-image-configuration

suggest: ```**Layer**: a set of filesystem changes; Layers are typically used by a layer-based or union filesystem to form the mountable root file system of a container.``` and maybe link to the oci spec

spec.md
```diff
@@ -5,76 +5,56 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
+- **Config**: a special layer defined at the top of a manifest containing artifact metadata
+- **Blob**: a single binary content stored in a registry
+- **Digest**: a unique blob identifier, calculated from the blob's content
+- **Content**: a general term for content that can be downloaded from a registry (manifest or blob), and are addressable by a digest
```
mikebrow, line 57: content is being used as a noun to represent a great many things throughout the spec.. in each case the thing being represented is or should be specified by the context.  

Suggest removing this definition since what is meant by it's use is really just the dictionary definition. Or just use the dictionary definition:
1a : something contained —usually used in plural
the layer's contents
the manifest's contents
the contents of the http response...

mikebrow: maybe a picture.. would help

spec.md
```diff
@@ -5,76 +5,56 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
 - [moby/moby#8093](https://github.com/moby/moby/issues/8093)
 - [moby/moby#9015](https://github.com/moby/moby/issues/9015)
 - [docker/docker-registry#612](https://github.com/docker/docker-registry/issues/612)
 
-<!--- TODO: add relevant background information here --->
+### Definitions
+
+Several terms are used frequently in this document and warrant basic definitions:
+
+- **Registry**: a HTTP service which implements this spec
```
mikebrow, line 47: s/HTTP//

spec.md
```diff
@@ -5,76 +5,56 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
+- **Client**: a tool that communicates with registries over HTTP
```
mikebrow, line 48: s/over HTTP//

spec.md
```diff
@@ -5,76 +5,56 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
+	- [Minimum Requirements](#minimum-requirements)
+	- [Official Certification](#official-certification)
+	- [Workflow Categories](#workflow-categories)
+		1. [Pull](#pull)
+		2. [Push](#push)
+		3. [Content Discovery](#content-discovery)
+		4. [Content Management](#content-management)
+- [HTTP API](#http-api)
+	- [Endpoints](#endpoints)
+	- [Error Codes](#error-codes)
```
vbatts, line 25: rebase for the prior TOC updates?

spec.md
dmcgowan: I think we should avoid using the term layers except in the context of images. I wish we would have just renamed layers to references. I don't think we need to inject that confusion here.

spec.md
vbatts: `a service which implements the server side of this spec`? because technically the client implements the spec too

spec.md
jdolitsky: rebased

spec.md
mikebrow: fair.. in the context of images image layers are a set of..

spec.md
jzelinskie: >a service that handles the required APIs in this specification

spec.md
jzelinskie: >a tool that consumes the APIs defined in this specification

spec.md
jzelinskie: I don't think the image spec is useful in this context. I'd rather define Layer as legacy terminology referring to blobs that are specifically for container images.

mikebrow: what is "this context?" ... Why is OCI image not useful in the context of the words used in the OCI distribution spec? Let's say an OCI image is but one type of OCI artifact... Most definitions include examples, for example, in the context of OCI container images, layers are a set of ...  IMO the inclusion of examples does not make a definition as confusing as the lack of examples does. Esp. when highjacking a term that is common in the vernacular.

spec.md
mikebrow: handles (the server side of) ? or is that repetitive?

spec.md
jzelinskie: A blob is the binary form of metadata OR [most commonly] content that is stored by the registry.
An artifact is one conceptual piece of content stored as blobs with an accompanying manifest containing metadata.
A layer is one kind of blob used for container image artifacts

rchincha: @jzelinskie vaguely remember you had a carefully compiled set of definitions. Let's look at that and see if we can just reuse that prose.

jdolitsky:
@jzelinskie @mikebrow can I please have a second review? I've updated the definitions based on your comments.

The following terms have been removed:

- Content
- Layer

The following terms are unchanged:

- Manifest
- Digest

The following terms have been updated:

- **Registry**: a service that handles the required APIs defined in this specification
- **Client**: a tool that communicates with registries
- **Push**: the act of uploading blobs and manifests to a registry
- **Pull**: the act of downloading blobs and manifests from a registry
- **Blob**: the binary form of content that is stored by a registry, addressable by a digest
- **Config**: a section in the manifest which contains artifact metadata
- **Artifact**: one conceptual piece of content stored as blobs with an accompanying manifest containing a config

jdolitsky:
CC @dmcgowan

spec.md
```diff
@@ -40,14 +40,19 @@ For relevant details and a history leading up to this specification, please see
...
+
+- **Registry**: a service that handles the required APIs defined in this specification
+- **Client**: a tool that communicates with registries
+- **Push**: the act of uploading blobs and manifests to a registry
+- **Pull**: the act of downloading blobs and manifests from a registry
+- **Blob**: the binary form of content that is stored by a registry, addressable by a digest
+- **Manifest**: a JSON document which defines an artifact
+- **Config**: a section in the manifest (and associated blob) which contains artifact metadata
+- **Artifact**: one conceptual piece of content stored as blobs with an accompanying manifest containing a config
+- **Digest**: a unique blob identifier, calculated from the blob's content
```
mikebrow, line 55: suggest: "a unique content identifier, for example: a hash calculated from a blob's content"

dmcgowan: I think it is worth mentioning this is a cryptographic hash. The digest is an important security aspect. Maybe just call it a "a unique identifier created from the cryptographic hash of a blob's content"

jdolitsky: Changed to Derek's suggestion

PR #166, opened 2020-06-16
mini update 6: describe pull workflow

pmengelbert:
builds upon PR #162 .

Here is the diff between this commit and PR #162:
https://github.com/bloodorangeio/distribution-spec/compare/mini-update-5...bloodorangeio:mini-update-6?expand=1

PR #167, opened 2020-06-16
mini update 7: describe push workflow

pmengelbert:
Builds on PR #166 

The relevant diff can be found here:
https://github.com/bloodorangeio/distribution-spec/compare/mini-update-6...bloodorangeio:mini-update-7?expand=1

This is a larger section and will need review before finalization--thank you for your comments.


rchincha:
The various flavors of Push I would imagine are related to the various client constraints one may have. For example, when memory constrained, when network is flaky (so need frequent resuming), when there are no real constrains (so I can push in one shot), etc. Clients today for whatever reason have decided to implement only a single flavor.

spec.md
```diff
@@ -79,11 +79,212 @@ Registry providers can self-certify by submitting conformance results to [openco
...
+Content-Type: application/vnd.oci.image.manifest.v1+json
+```
+```
+<manifest byte stream>
+```
+
+`<name>` is the namespace of the repository, and the `<reference>` MUST be either a) a digest or b) a tag.
+```regexp
+^[0-9]+-[0-9]+$
+```
```
jdolitsky, line 275: I think this can be removed

spec.md
```diff
@@ -79,11 +79,212 @@ Registry providers can self-certify by submitting conformance results to [openco
...
+and body:
+`/v2/<name>/manifests/<reference>`
+```http request
+Content-Type: application/vnd.oci.image.manifest.v1+json
+```
+```
+<manifest byte stream>
+```
+
+`<name>` is the namespace of the repository, and the `<reference>` MUST be either a) a digest or b) a tag.
```
jdolitsky, line 272: nothing actionable in this PR, but tag should be added to definitions

jdolitsky:
Closing this in favor of https://github.com/opencontainers/distribution-spec/pull/178

Issue #68, opened 2019-08-14
Review status codes for 1.0


dmcgowan:
We need to go through and review all the returned status codes for the various endpoints. This includes what is defined in the overview, what is defined in the detail, and whether it tested and which status code the test uses.

The result of this should be removing the disclaimer about discrepancies between header and detail.

| API endpoint | Overview | Detail | Tests | Status |
|---|---|---|---|---|
| `GET /v2/` | `200` | `200` | | |
| `GET /v2/<name>/tags/list` |  | `200` | |
| `GET /v2/<name>/tags/list?n=<integer>&last=<integer>` | | `200` | |
| `GET /v2/<name>/manifests/<reference>` | | `200` | |
| `PUT /v2/<name>/manifests/<reference>` | | `201` | |
| `DELETE /v2/<name>/manifests/<reference>` | | `202` | |
| `GET /v2/<name>/blobs/<digest>` | | `200` | |
| `GET /v2/<name>/blobs/<digest>` + `Range` | | `206` | |
| `DELETE /v2/<name>/blobs/<digest>` | [`202`](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#deleting-a-layer) | [`202`](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#delete-blob) | | Inconsistent wording (layer vs blob) |
| `POST /v2/<name>/blobs/uploads/?digest=<digest>` | | `201` | |
| `POST /v2/<name>/blobs/uploads/` | | `202` | |
| `POST /v2/<name>/blobs/uploads/?mount=<digest>&from=<repository name>` | | `201` | |
| `GET /v2/<name>/blobs/uploads/<uuid>` | [`204`](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#upload-progress) | [`204`](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#get-blob-upload) | |
| `PATCH /v2/<name>/blobs/uploads/<uuid>` | [`202`](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#chunked-upload) | [`202`](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#patch-blob-upload) | |
[...]

PR #171, opened 2020-07-10
Expand allowed status codes

pmengelbert:
POST /v2/\<name\>/blobs/uploads/?digest=\<digest\> should allow for 202 Accepted.
All DELETE calls should allow 405 Method Not Allowed.

This PR resolves the following comment by @jdolitsky on [issue 68](https://github.com/opencontainers/distribution-spec/issues/68#issuecomment-628254410_). Resolving this issue will make several registries go green on the relevant workflows.

### DIFF
--- a/conformance/02_push_test.go
+++ b/conformance/02_push_test.go
@@ -71,7 +71,10 @@ var test02Push = func() {
 				Expect(err).To(BeNil())
 				location := resp.Header().Get("Location")
 				Expect(location).ToNot(BeEmpty())
-				Expect(resp.StatusCode()).To(Equal(http.StatusCreated))
+				Expect(resp.StatusCode()).To(SatisfyAny(
+					Equal(http.StatusCreated),
+					Equal(http.StatusAccepted),
+				))
 			})
 
 			g.Specify("GET request to blob URL from prior request should yield 200", func() {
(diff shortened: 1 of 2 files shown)


jdolitsky:
LGTM

jdolitsky:
Should be noted that the current spec is not as permissive, but it will be factored into the rewrite

vbatts:
glad we're testing for this.
I recall the discussion for 202 being used in addition to the [201 written in the spec](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#post-initiate-blob-upload).
Is there other background on using `202 Accepted` that should be linked here? or update to the spec needed to relate to this conformance test?

jdolitsky:
@vbatts its something that should be accounted for in the spec reorg: https://github.com/opencontainers/distribution-spec/pull/178

pmengelbert:
@vbatts This has been included in the latest commit for PR #178

vbatts:
the maintainers had chatted about this in a call recently. While everyone noted that some of these response codes are being used just to mimic docker registry, they also noted that they really ought to be corrected in the original registry implementation.

vbatts:
I'm not opposed to this conformance test, but might we track the conversation to have this aligned in the community, rather than only widening the conformance to make it green for more registries?

jdolitsky:
FWIW this change is now reflected in the spec.md on master branch

vbatts:
@jdolitsky are saying that this PR is no longer needed? or just that the spec has the language supporting this?

jdolitsky:
@vbatts - the spec already contains language, so this PR should prob be merged

jdolitsky:
closing this. we settled on final HTTP codes for 1.0

PR #178, opened 2020-08-11
Reorganize distribution spec

pmengelbert:
* New spec.md
* Move 'detail' section into detail.md

To view the revised spec in complete form, see the following:
https://github.com/bloodorangeio/distribution-spec/blob/new-spec/spec.md


spec.md
```diff
@@ -5,4599 +5,421 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
-| `WWW-Authenticate` | An RFC7235 compliant authentication challenge header. |
-| `Content-Length`   | Length of the JSON response body.                     |
+`<name>` is the namespace of the repository, and `<int>` is an integer specifying the number of tags requested. The response
+to such a request MAY return fewer than `<int>` results, but only when the total number of tags attached to the repository
+is less than `<int>`. Otherwise, the response MUST include `<int>` results. Without the `last` query parameter (described
+next), the list returned will start at the beginning of the list and include `<int>` results. As above, the tags MUST be
+in lexical order.
 
-The error codes that MAY be included in the response body are enumerated below:
+The `last` query parameter provides further means for limiting the number of tags. It is used exclusively in combination with the
```
hallyn, line 322: 'last' seems an odd name.  Wouldn't something like 'startafter' be more correct?

spec.md
```diff
@@ -5,4599 +5,421 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
+To fetch the list of tags, perform a `GET` request to a path in the following format:
+[8a](#Endpoints) `/v2/<name>/tags/list`
 
-```HTTP
-401 Unauthorized
-WWW-Authenticate: <scheme> realm="<realm>", ..."
-Content-Length: <length>
-Content-Type: application/json
+`<name>` is the namespace of the repository. Assuming a repository is found, this request MUST return a
+`200 OK` response code. The list of tags MAY be empty, if there are no tags on the repository. If the list is not empty,
```
hallyn, line 295: nit- maybe drop the comma after 'empty'?

spec.md
```diff
@@ -5,4599 +5,421 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
-
-The following headers will be returned on the response:
-
-| Name             | Description                       |
-|------------------|-----------------------------------|
-| `Content-Length` | Length of the JSON response body. |
+##### Deleting tags
+`<name>` is the namespace of the repository, and `<tag>` is the name of the tag to be deleted. Upon success, the registry
+MUST respond with a `202 Accepted` code. If tag deletion is disabled, the registry MUST respond with either a
+`400 Bad Request` or a `405 Method Not Allowed`.
```
hallyn, line 338: Does the spec want to take a stand on auto-deletion of a manifest when the last tag to it is removed? In particular, should that be disallowed? Else clients may have to try and guess what a repo will do...

spec.md
```diff
@@ -5,4599 +5,421 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
-    "tags": [
-        "<tag>",
-        ...
-    ]
-}
+```json
+    {
+        "errors": [
+            {
+                "code": "<error identifier>",
```
hallyn, line 394: Should this say "(see below)" ?

spec.md
```diff
@@ -5,4599 +5,421 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
-This specification includes the following features:
+- **Registry**: a service that handles the required APIs defined in this specification
+- **Client**: a tool that communicates with registries
+- **Push**: the act of uploading blobs and manifests to a registry
+- **Pull**: the act of downloading blobs and manifests from a registry
+- **Blob**: the binary form of content that is stored by a registry, addressable by a digest
+- **Manifest**: a JSON document which defines an artifact. Manifests are defined under the [OCI Image Spec](https://github.com/opencontainers/image-spec/blob/master/manifest.md)
+- **Config**: a section in the manifest (and associated blob) which contains artifact metadata
+- **Artifact**: one conceptual piece of content stored as blobs with an accompanying manifest containing a config
+- **Digest**: a unique identifier created from a cryptographic hash of a blob's content. Digests are defined under the [OCI Image Spec](https://github.com/opencontainers/image-spec/blob/b6e51fa50549ee0bd5188494912a7f4c382cb0d4/descriptor.md#digests)
```
hallyn, line 55: Is that a commit id in the link? That will quickly make itself obsolete won't it?

spec.md
```diff
@@ -5,4599 +5,421 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
-If process A and B upload the same layer at the same time, both operations will proceed and the first to complete will be stored in the registry (Note: we MAY modify this to prevent dogpile with some locking mechanism).
+To pull a blob, perform a `GET` request to a url in the following form:
+[2a](#Endpoints) `/v2/<name>/blobs/<digest>`
 
-## Changes
+`<name>` is the namespace of the repository, and `<digest>` is the blob's digest.
 
-The V2 specification has been written to work as a living document, specifying only what is certain and leaving what is not specified open or to future changes.
-Only non-conflicting additions SHOULD be made to the API and accepted changes SHOULD avoid preventing future changes from happening.
+A GET request to an existing blob URL MUST provide the expected blob, with a reponse code that MUST be `200 OK`.
```
hallyn, line 90: Ignorant question - is there any question of permissions here?

spec.md
```diff
@@ -5,4599 +5,421 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
+To pull a manifest, perform a `GET` request to a url in the following form:
+[3a](#Endpoints) `/v2/<name>/manifests/<reference>`
 
-```HTTP
-/v2/<name>/
-```
+`<name>` refers to the namespace of the repository. `<reference>` MUST be either (a) the digest of the manifest or (b) a tag name.
 
-For example, an API endpoint that will work with the `library/ubuntu` repository, the URI prefix will be:
+The `<reference>` MUST NOT be in any other format.
```
hallyn, line 101: Seems like the two sentences referring to 'reference' should be in the same paragraph?

spec.md
pmengelbert: This is how the prior spec was written, and how all registries currently operate. I agree that 'startafter' might be more correct, but the proposed v1 spec is more descriptive than prescriptive, and we don't want the spec to be out of sync with how all registries currently operate.

spec.md
pmengelbert: :+1:

spec.md
pmengelbert: We are avoiding discussion of the authentication/authorization system. There is a "Docker-style" system, but it is not under the purview of this spec. The response codes assume successful authorization.

spec.md
pmengelbert: That's helpful, yes

spec.md
pmengelbert: Good call -- I'll update with a reference to master

spec.md
pmengelbert: The spec doesn't take a stand on this, and registries implement a fairly wide range of garbage-collecting behavior.  It is worth opening an issue to discuss, possibly for inclusion in a post-1.0 spec.

pmengelbert: I opened the issue here: #180

detail.md
```diff
@@ -0,0 +1,3761 @@
+## Detail
+
+> **Note**: This section is still under construction.
+> For the purposes of implementation, if any details below differ from the described request flows above, the section below SHOULD be corrected.
+> When they match, this note SHOULD be removed.
+
+The behavior of the endpoints are covered in detail in this section, organized by route and entity.
```
mikebrow, line 7: route and entity? 

We should define route and entity before using them here as the main introduction for the section I think.
`route: API protocol flow consisting of requests and responses between two entities implementing the API. `

spec.md
```diff
@@ -5,4599 +5,420 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
-> __IMPORTANT:__ If a _digest_ is used to fetch content, the client SHOULD use the same digest used to fetch the content to verify it.
-> The header `Docker-Content-Digest` SHOULD NOT be trusted over the "local" digest.
+Successful completion of the request MUST return either a `201 Created` or a `202 Accepted`, and MUST include the following header:
 
-### Pulling An Image
-
-An "image" is a combination of a JSON manifest and individual layer files.
-The process of pulling an image centers around retrieving these two components.
-
-The first step in pulling an image is to retrieve the manifest.
```
mikebrow, line 295: ..

spec.md
```diff
@@ -5,4599 +5,420 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
-
-This endpoint MAY issue a 307 (302 for < HTTP 1.1) redirect to another service for downloading the layer and clients SHOULD be prepared to handle redirects.
-
-This endpoint SHOULD support aggressive HTTP caching for image layers.
-Support for Etags, modification dates and other cache control headers SHOULD be included.
-To allow for incremental downloads, `Range` requests SHOULD be supported, as well.
-
-### Pushing An Image
-
-Pushing an image works in the opposite order as a pull.
```
mikebrow, line 383: ..

spec.md
```diff
@@ -5,4599 +5,420 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
-After assembling the image manifest, the client MUST first push the individual layers.
-When the layers are fully pushed into the registry, the client SHOULD upload the signed manifest.
-
-The details of each step of the process are covered in the following sections.
-
-#### Pushing a Layer
-
-All layer uploads use two steps to manage the upload process.
```
mikebrow, line 391: ..

spec.md
```diff
@@ -5,4599 +5,420 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
 
-For detail on individual endpoints, please see the [_Detail_](#detail) section.
+There are two ways to push a blob monolithically:
+1. A single `POST` request
+2. A `POST` request followed by a `PUT` request
 
-### Errors
+---
 
-Actionable failure conditions, covered in detail in their relevant sections, are reported as part of 4xx responses, in a json response body.
```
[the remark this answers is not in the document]
jdolitsky: This should be covered in the "Error Codes" section

detail.md
pmengelbert: Done

spec.md
```diff
@@ -5,4599 +5,420 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
-If a `401 Unauthorized` response is returned, the client SHOULD take action based on the contents of the "WWW-Authenticate" header and try the endpoint again.
-Depending on access control setup, the client MAY still have to authenticate against different resources, even if this check succeeds.
-
-If `404 Not Found` response status, or other unexpected status, is returned, the client SHOULD proceed with the assumption that the registry does not implement V2 of the API.
-
-When a `200 OK` or `401 Unauthorized` response is returned, the "Docker-Distribution-API-Version" header SHOULD be set to "registry/2.0".
-Clients MAY require this header value to determine if the endpoint serves this API.
-When this header is omitted, clients MAY fallback to an older API version.
-
-### Content Digests
```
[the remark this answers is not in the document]
pmengelbert: Restored to content-addressability.md

spec.md
```diff
@@ -5,4599 +5,420 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
 
-This scheme provides rich access control over various operations and methods using the URI prefix and http methods that can be controlled in variety of ways.
+If the manifest is not found in the registry, the response code MUST be `404 Not Found`.
 
-Classically, repository names have always been two path components where each path component is less than 30 characters.
-The V2 registry API does not enforce this.
-The rules for a repository name are as follows:
+#### Push
 
-1. A repository name is broken up into _path components_.
```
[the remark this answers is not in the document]
pmengelbert: It's been replaced with a description of the whole repository name, including the original regex used to specify the pattern that must be matched.

spec.md
```diff
@@ -5,4599 +5,420 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
+- **Config**: a section in the manifest (and associated blob) which contains artifact metadata
+- **Artifact**: one conceptual piece of content stored as blobs with an accompanying manifest containing a config
+- **Digest**: a unique identifier created from a cryptographic hash of a blob's content. Digests are defined under the [OCI Image Spec](https://github.com/opencontainers/image-spec/blob/master/descriptor.md)
+- **Tag**: a custom, human-readable manifest identifier
 
-The following is an incomplete list of features, discussed during the process of cutting this specification, which MAY be out of the scope of this specification, MAY be the purview of another specification, or MAY be deferred to a future version:
+## Conformance
 
-- Authentication and authorization support: While authentication and authorization support will influence this specification, those details MAY be left to a future specification. However, relevant header definitions and error codes are present to provide an indication of what a client MAY encounter.
```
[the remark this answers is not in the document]
pmengelbert: Thanks -- we have restored the entire "Scope" section

spec.md
```diff
@@ -5,4599 +5,420 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
-- Migration from v2compatibility representation
+### Notational Conventions
 
-## Use Cases
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119) (Bradner, S., "Key words for use in RFCs to Indicate Requirement Levels", BCP 14, RFC 2119, March 1997).
 
-### Image Verification
+### Requirements
 
-A container engine would like to run verified image named "library/ubuntu", with the tag "latest".
```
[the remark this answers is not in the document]
pmengelbert: We moved the Use Cases section to scenarios.md

spec.md
pmengelbert: Strictly speaking, a layer upload can be done in a single step: a POST request containing the blob with the ?digest=<digest> query parameter will complete an upload in a single-shot fashion.

spec.md
pmengelbert: Restored

spec.md
pmengelbert: Restored

pmengelbert:
@mikebrow Let me know if all of your concerns have been addressed. Thanks for the review!

vbatts:
What a monumental effort. The miniupdate branch was the best way to break this rewrite up.
Doing a once-through looks essentially all there.
Thanks @jdolitsky and @pmengelbert for this work.

LGTM

spec.md
```diff
@@ -97,4507 +77,377 @@ The following is an incomplete list of features, discussed during the process of
...
-        ]
-    }
-```
+`<name>` refers to the namespace of the repository. `<reference>` MUST be either (a) the digest of the manifest or (b) a tag name.
+The `<reference>` MUST NOT be in any other format.
 
-The `code` field will be a unique identifier, all caps with underscores by convention.
-The `message` field will be a human readable string.
-The OPTIONAL `detail` field MAY contain arbitrary json data providing information the client can use to resolve the issue.
+A GET request to an existing manifest URL MUST provide the expected manifest, with a response code that MUST be `200 OK`.
```
mikebrow, line 117: expected.. is pretty broad... before ending this section should make a statement regarding how the client should use the response to validate the manifest .. and/or link to the detail doc for the associated APIs.

Is the idea that the reader will hop back and forth between this doc and detail? Or is this document supposed to stand on it's own?

This document seems to be more of an outline now ... with an intro to the workflows. We should probably not have MUST language in the rough description of each flow.  Otherwise the implementor will feel like they have to check two places with two different descriptions of the APIs to see what they MUST implement for each.

scenarios.md
```diff
@@ -0,0 +1,30 @@
+## Use Cases
+
+### Image Verification
```
mikebrow, line 3: TODO: link/weave these scenarios into the overall spec somehow

spec.md
```diff
@@ -97,4507 +77,377 @@ The following is an incomplete list of features, discussed during the process of
...
+| `MANIFEST_INVALID`      | manifest invalid                               |
+| `MANIFEST_UNKNOWN`      | manifest unknown                               |
+| `MANIFEST_UNVERIFIED`   | manifest failed signature verification         |
+| `NAME_INVALID`          | invalid repository name                        |
+| `NAME_UNKNOWN`          | repository name not known to registry          |
+| `SIZE_INVALID`          | provided length did not match content length   |
+| `TAG_INVALID`           | manifest tag did not match URI                 |
+| `UNAUTHORIZED`          | authentication required                        |
+| `DENIED`                | requested access to the resource is denied     |
+| `UNSUPPORTED`           | the operation is unsupported                  |
```
mikebrow, line 453: doc ends pretty abruptly ... link to detail? add references ?

spec.md
```diff
@@ -5,76 +5,56 @@ keywords: registry, on-prem, images, tags, repository, distribution, api, advanc
...
+
+- **Registry**: a service that handles the required APIs defined in this specification
+- **Client**: a tool that communicates with registries
+- **Push**: the act of uploading blobs and manifests to a registry
+- **Pull**: the act of downloading blobs and manifests from a registry
+- **Blob**: the binary form of content that is stored by a registry, addressable by a digest
+- **Manifest**: a JSON document which defines an artifact. Manifests are defined under the [OCI Image Spec](https://github.com/opencontainers/image-spec/blob/master/manifest.md)
+- **Config**: a section in the manifest (and associated blob) which contains artifact metadata
+- **Artifact**: one conceptual piece of content stored as blobs with an accompanying manifest containing a config
+- **Digest**: a unique identifier created from a cryptographic hash of a blob's content. Digests are defined under the [OCI Image Spec](https://github.com/opencontainers/image-spec/blob/master/descriptor.md)
```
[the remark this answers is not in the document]
jdolitsky: i.e. link to new content-addressibility.md vs image spec?

spec.md
jdolitsky: We feel the document stands on its own (or should).

Regarding validation of manifest etc., that seems to be something that should be discussed in the spec which defines it (image spec).

We were ready to do without the details section, but added it per maintainers' request. Ultimately the conformance section should be everything needed to implement the spec.

spec.md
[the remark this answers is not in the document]
mikebrow: both.. i suppose
or todo to resolve how the new file should be used in this spec.

jdolitsky:
@vbatts - we appreciate this!

@mikebrow opened https://github.com/opencontainers/distribution-spec/issues/181 to link back and forth between the new spec and these other docs (content-addressibility.md etc.)

Like you said, I think it's best to get this merged then follow up with tiny PRs to amend prior to 1.0

spec.md
mikebrow: >  "stands on its own (or should)" ... "conformance section should be everything needed to implement the spec"

Given the answer to the image digest/signature, I presume you mean with respect to the details document that you don't want there, not with respect to other OCI spec docs. I'm against having everything needed to implement the distribution spec APIs in "two" separate documents, esp. where the MUST language of the APIs is concerned.

> "should be discussed in the spec which defines it (image spec)" 

The OCI image spec does define this as a MUST requirement and further specifies that "retrieved content SHOULD be verified against this digest when consumed via untrusted sources."  I should hope the same will be true for all artifact types.  We are talking about the "when consumed part."  What do you think about using some general language here, for all the workflows if you like, for suggesting that "verify" step(s) MUST be provided according to the requirements and suggestions made by the specifications for the media type being distributed, IE for OCI manifests, blobs.. etc.. A simple example can be included for pull, but I would agree you don't have to say the same thing on every workflow/ every API.

mikebrow: summary resolution:
TODO: add general text about artifact validation requirements 

goal will be to make spec.md standalone from details.md and we me be able to remove most of the contents of details.md

jdolitsky:
LGTM

mikebrow:
thx guys! tty on monday to begin next phase :-)

PR #192, opened 2020-10-01
Add go.mod

ashokponkumar:
Signed-off-by: Ashok Pon Kumar <ashokponkumar@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/opencontainers/distribution-spec
+
+go 1.15
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.13.x
+  - 1.15.x
 
 sudo: required
 


jdolitsky:
@ashokponkumar was this required for you to use some of the Go libs in this repo?

ashokponkumar:
> @ashokponkumar was this required for you to use some of the Go libs in this repo?

Not for this repo specifically. I am using the runtime-spec in https://github.ibm.com/app-modernization/move2kube/blob/master/go.mod and I noticed that some of the other spec's were missing the go.mod and that it might create a problem once we hit v2, hence added them and raised the pull request.

go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/opencontainers/distribution-spec
```
mikebrow, line 1: Should this be based off `go mod init specs-go/specs` As there are no go files in the root or packages..

resulting in `module specs-go/specs`

go.mod
```diff
@@ -0,0 +1,3 @@
...
+
+go 1.14
```
mikebrow, line 3: should either move up travis to 1.14 or 15 or change this to go 1.13 to match travis test.. See: https://github.com/opencontainers/distribution-spec/blob/master/.travis.yml#L3

mikebrow (commented):
see questions comments...

go.mod
mikebrow: have you tested it over in another repo? I didn't try this out..

ashokponkumar:
Thanks @mikebrow for the comments. @thaJeztah has created a proposal for a similar PR in runtime-spec in https://github.com/opencontainers/runtime-spec/issues/1069. Based on how that proposal evolves, I can appropriately make the changes here.

go.mod
ashokponkumar: Sure @mikebrow. Have moved both to 1.15

go.mod
ashokponkumar: We can evolve it similar to https://github.com/opencontainers/runtime-spec/issues/1069

vbatts:
Is this still needed?

thaJeztah:
I think this is still desirable, but some discussion needed on the approach (`go.mod` in subdirectory, not at the top, and potentially separate versioning to allow fixes that only affect the go-code, not the spec itself); https://github.com/opencontainers/distribution-spec/pull/192#issuecomment-702513616
 
> Thanks @mikebrow for the comments. @thaJeztah has created a proposal for a similar PR in runtime-spec in [opencontainers/runtime-spec#1069](https://github.com/opencontainers/runtime-spec/issues/1069). Based on how that proposal evolves, I can appropriately make the changes here.

vbatts (approved):
🎇

vbatts:
maybe better if someone carries this and put's the go.mod in the right place

vbatts:
closing in favor of #283

PR #203, opened 2020-10-26
fixed to use "bytes" unit for Content-Range to spec/test

Code-Hex:
According [rfc7233](https://tools.ietf.org/html/rfc7233#page-12), We need "bytes" prefix.

Grammar is here

```
 byte-content-range  = bytes-unit SP
                           ( byte-range-resp / unsatisfied-range )
```

### DIFF
--- a/conformance/setup.go
+++ b/conformance/setup.go
@@ -246,10 +246,10 @@ func init() {
 	testBlobBDigest = godigest.FromBytes(testBlobB).String()
 	testBlobBChunk1 = testBlobB[:3]
 	testBlobBChunk1Length = strconv.Itoa(len(testBlobBChunk1))
-	testBlobBChunk1Range = fmt.Sprintf("0-%d", len(testBlobBChunk1)-1)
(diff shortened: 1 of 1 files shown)


pmengelbert:
@Code-Hex I tested this fix against three different registries, and it caused all of them to fail the conformance tests. I appreciate the attention to detail in conforming to RFC 7233, but this spec is derived from the original Docker distribution specification, which itself was out of sync with that RFC.

Because of this, most registries do not expect the 'bytes' prefix, and adding it breaks their conformance. This would require most registries to code a solution, which is not ideal, especially so shortly before the 1.0 release.

@jdolitsky

jdolitsky:
+1. I'm not seeing any reference to RFC7233 in spec. Going to drop fro rc2 milestone

jdolitsky:
although it would be better in theory to follow the RFC, it was discussed and it doesn't make sense to break people for this.

Please resubmit into spec.md with language vs. conformance code changes if you believe this should be optional behaviour

PR #204, opened 2020-10-27
Add cross-repository mounting section to spec/test

pmengelbert:
* New test in 02_push_test.go
* Section added to spec.md detailing cross-repository mounting
* Section added to conformance/README.md on new configuration options
for cross-repository mounting

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -321,6 +321,8 @@ Location: <blob-location>
 The digest contained in the `Location` header MAY be different from that of the blob that was mounted. As such, a client
 SHOULD use the digest found in the path from this header and SHOULD NOT use the digest of the blob that was mounted.
 
+If the source blob is not found, the response MUST be `404 Not Found`.
+
 ##### Pushing Manifests
 
 To push a manifest, perform a `PUT` request to a path in the following format, and with the following headers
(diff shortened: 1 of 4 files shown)


jdolitsky (approved):
LGTM. Tested against Docker distribution

spec.md
```diff
@@ -304,6 +304,23 @@ Location: <blob-location>
...
+##### Mounting a blob from another repository
+If a necessary blob exists already in another repository, it can be mounted into a different repository via a `POST`
+request in the following format:
+
+`/v2/<name>/blobs/uploads/?mount=<digest>&from=<other_namespace>`  <sup>[end-11](#endpoints)</sup>.
+
+In this case, `<name>` is the namespace to which the blob will be mounted. `<digest>` is the digest of the blob to mount,
+and `<other_namespace>` is the namespace from which the blob should be mounted.
+
+The response to a successful mount MUST be `201 Created`, and MUST contain the following header:
```
dmcgowan, line 316: I think it should be worth mentioning that this is normally done in place of the normal call to `.../uploads/` to start a blob upload. So if a registry does not support cross repository mount, it will return a normal `202` and the client will continue the upload (not reissue the request without the mount param).

Did we have `201` listed as valid return code for `/uploads` without mount for a reason? Or was that just listed because of this. We should be explicit about the difference between `201` and `202` here.

pmengelbert: @dmcgowan based on yesterday's call, we have decided to require cross-repository mount, making the distinction between `201` and `202` unnecessary here -- it will always be `201` unless the source blob is not found, in which case it will be `404`.

Allowing both `201` and `202` on `/uploads` was a compromise solution based on what existing registries were already doing -- see #171 and  @jdolitsky 's comment on #68

jonjohnsonjr: I've written several clients that rely on the fact that `201` means the mount succeeded and `202` means continue with the upload -- ideally registry behavior would be consistent here. I don't think requiring cross-repo mount invalidates the requirement of this distinction.

Wwwsylvia:
LGTM. Tested against ACR.

spec.md
jonjohnsonjr: > unless the source blob is not found, in which case it will be `404`

This would also slow down uploads because you'd require an additional roundtrip to start the upload process if the mount fails.

pmengelbert: @jonjohnsonjr This is my proposed solution:
In the first test, allow either a `201` or `202`. If `201`, a fetch for the blob will be tested to ensure its existence. If `202`, we will verify that a valid `<session-id>` is returned.

If I'm getting it right, this should satisfy the requirements -- please let me know if there's anything further I should know about before making changes.

pmengelbert:
@jonjohnsonjr @dmcgowan I just pushed the requested changes. Please review, and feel free to approve if all looks good. Thanks!

Issue #202, opened 2020-10-15
spec intro a little tighter


SteveLasker:
> The Open Container Initiative Distribution Specification (a.k.a. "OCI Distribution Spec") defines an API protocol to facilitate and standardize the distribution of content.
> While this specification is designed to be agnostic to content types, much of it is centered around the distribution of container images. Many of the concepts here, such as "manifests" and "digests", are originally defined in the Open Container Initiative Image Format Specification (a.k.a. "OCI Image Spec"), and the OCI Image is considered to be the primary supported artifact type.
> For guidance on how to apply this specification to other artifact types, please see the Open Container Initiative Artifact Authors Guide (a.k.a. "OCI Artifacts").

The spec is certainly originally focused on containers. As we've seen, there are already more uses cases, prior to OCI Artifacts. We've also seen the distribution spec is pretty vague about what it actually stores as none of the `mediaTypes`, nor even the sequence of events for pushing content (layers, config, manifest) are defined. 

The intro is a little more generic, but can I suggest the following:

---

_First paragraph remains the same:_
> The Open Container Initiative Distribution Specification (a.k.a. "OCI Distribution Spec") defines an API protocol to facilitate and standardize the distribution of content.

While OCI Image is the most prominent, the specification is designed to be agnostic of content types. Concepts such as "manifests" and "digests", are currently defined in the [Open Container Initiative Image Format Specification](https://github.com/opencontainers/image-spec) (a.k.a. "OCI Image Spec").

To support other artifact types, please see the [Open Container Initiative Artifact Authors Guide](https://github.com/opencontainers/artifacts) (a.k.a. "OCI Artifacts").

jdolitsky:
Resolved by #206

Issue #201, opened 2020-10-14
Conformance: test for cross-repo blob mounting


jdolitsky:
related to #186

jdolitsky:
Resolved by #206

Issue #195, opened 2020-10-09
Config blob definition misleading?


hallyn:
Maybe it's my reading of it, or my misunderstanding of the possibilities,
but I've always seen a Config section in a Manifest to be a pointer to the
config blog.  But the definitions here say:

```
Manifest: a JSON document which defines an Artifact. Manifests are defined under the OCI Image Spec apdx-2
Config: a section in the Manifest (and associated Blob) which contains Artifact metadata
```

It would seem more accurate to me to say that a Config is a blob, referenced
from a Manifest, describing that manifest.

jdolitsky:
> It would seem more accurate to me to say that a Config is a blob, referenced
from a Manifest, describing that manifest.

Yes :)

jdolitsky:
Resolved by #206

Issue #186, opened 2020-09-18
Spec seems to be missing cross-repo mounting entirely


jonjohnsonjr:
This section from the docker spec seems to have been dropped: https://github.com/docker/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#cross-repository-blob-mount

I see that some info is still present in detail.md, which isn't referenced by spec.md at all: https://github.com/opencontainers/distribution-spec/blob/master/detail.md

Was this omission intentional?

jdolitsky:
I think the omission is intentional, as mounting isn't fully worked out for 1.0 release. Was discussed that this, along with extensions, and other topics can be reworked/re-added for subsequent releases.

It is intended that detail.md will be referenced in spec.md prior to 1.0

jonjohnsonjr:
> Was discussed that this, along with extensions, and other topics can be reworked/re-added for subsequent releases.

Interesting! Do you remember where the discussion took place? Have a link handy?

jdolitsky:
@jonjohnsonjr It was a vidchat several weeks ago between maintainers related to getting this PR merged: https://github.com/opencontainers/distribution-spec/pull/178

Since we didn't want to lose useful bits, big chunk was saved off into details.md, although it shouldn't be considered "official". Working on adding that language in #188 

@vbatts - correct me if I'm wrong about punting extensions stuff until after 1.0

dmcgowan:
> I think the omission is intentional, as mounting isn't fully worked out for 1.0 release.

I don't think that is accurate. Mounting is fully defined and regularly used today, I haven't heard of any aspect of it that was considered to be in-progress.

jdolitsky:
@dmcgowan ok, lets add it back. Which section should it be defined under?

dmcgowan:
Next to regular blob push. Used when initiating a new upload

jdolitsky:
@jonjohnsonjr @dmcgowan please see https://github.com/opencontainers/distribution-spec/pull/204

jdolitsky:
Resolved by #206

PR #206, opened 2020-10-28
Resolve several open issues

### LINKED ISSUE #193: Update "conformance" section
In the current generated HTML/PDF spec, "TODO: add general text about artifact validation requirements"

We should update that with general information on how to run the conformance test and point people to https://github.com/opencontainers/oci-conformance

pmengelbert:
Built on top of PR #204 

Resolves #193 #26 #60 #68 #79 #126 #132 #183 #186 #195 #201 #202

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -28,11 +28,10 @@
 
 The **Open Container Initiative Distribution Specification** (a.k.a. "OCI Distribution Spec") defines an API protocol to facilitate and standardize the distribution of content.
 
-While this specification is designed to be agnostic to content types, much of it is centered around the distribution of container images.
-Many of the concepts here, such as "manifests" and "digests", are originally defined in the [Open Container Initiative Image Format Specification](https://github.com/opencontainers/image-spec) (a.k.a. "OCI Image Spec"),
-and the OCI Image is considered to be the primary supported artifact type.
+While OCI Image is the most prominent, the specification is designed to be agnostic of content types. Concepts such as "manifests" and "digests",
+are currently defined in the [Open Container Initiative Image Format Specification](https://github.com/opencontainers/image-spec) (a.k.a. "OCI Image Spec").
 
-For guidance on how to apply this specification to other artifact types, please see the [Open Container Initiative Artifact Authors Guide](https://github.com/opencontainers/artifacts/blob/master/artifact-authors.md) (a.k.a. "OCI Artifacts").
+To support other artifact types, please see the [Open Container Initiative Artifact Authors Guide](https://github.com/opencontainers/artifacts) (a.k.a. "OCI Artifacts").
 
 ### Historical Context
 
@@ -59,7 +58,7 @@ Several terms are used frequently in this document and warrant basic definitions
 - **Pull**: the act of downloading Blobs and Manifests from a Registry
 - **Blob**: the binary form of content that is stored by a Registry, addressable by a Digest
 - **Manifest**: a JSON document which defines an Artifact. Manifests are defined under the OCI Image Spec <sup>[apdx-2](#appendix)</sup>
-- **Config**: a section in the Manifest (and associated Blob) which contains Artifact metadata
+- **Config**: a blob referenced in the Manifest (and associated Blob) which contains Artifact metadata and describes the Manifest
 - **Artifact**: one conceptual piece of content stored as Blobs with an accompanying Manifest containing a Config
 - **Digest**: a unique identifier created from a cryptographic hash of a Blob's content. Digests are defined under the OCI Image Spec <sup>[apdx-3](#appendix)</sup>
 - **Tag**: a custom, human-readable Manifest identifier
@@ -79,14 +78,14 @@ The engine contacts the registry, requesting the manifest for "library/ubuntu:la
 An untrusted registry returns a manifest.
 After each layer is downloaded, the engine verifies the digest of the layer, ensuring that the content matches that specified by the manifest.
 
-### Resumable [Push](#push)
+### Resumable Push
 
 Company X's build servers lose connectivity to a distribution endpoint before completing an artifact layer transfer.
 After connectivity returns, the build server attempts to re-upload the artifact.
 The registry notifies the build server that the upload has already been partially attempted.
 The build server responds by only sending the remaining data to complete the artifact file.
 
-### Resumable [Pull](#pull)
+### Resumable Pull
 
 Company X is having more connectivity problems but this time in their deployment datacenter.
 When downloading an artifact, the connection is interrupted before completion.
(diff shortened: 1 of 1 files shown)


jdolitsky (approved):

spec.md
```diff
@@ -54,7 +58,7 @@ Several terms are used frequently in this document and warrant basic definitions
 - **Pull**: the act of downloading Blobs and Manifests from a Registry
 - **Blob**: the binary form of content that is stored by a Registry, addressable by a Digest
 - **Manifest**: a JSON document which defines an Artifact. Manifests are defined under the OCI Image Spec <sup>[apdx-2](#appendix)</sup>
-- **Config**: a section in the Manifest (and associated Blob) which contains Artifact metadata
+- **Config**: a blob referenced in the Manifest (and associated Blob) which contains Artifact metadata and describes the Manifest
```
jonjohnsonjr, line 61: The `(and associated Blob)` here no longer makes sense.

> and describes the Manifest

Does it?

spec.md
```diff
@@ -304,6 +311,28 @@ Location: <blob-location>
 Here, `<blob-location>` is a pullable blob URL.
 
 
+##### Mounting a blob from another repository
+If a necessary blob exists already in another repository, it can be mounted into a different repository via a `POST`
+request in the following format:
+
+`/v2/<name>/blobs/uploads/?mount=<digest>&from=<other_namespace>`  <sup>[end-11](#endpoints)</sup>.
```
jonjohnsonjr, line 318: The original spec referred to this as `<repository name>` -- why change it?

There may have been other conversations about preferring namespace over repository that I missed. If so feel free to ignore me.

spec.md
pmengelbert: I adjusted the language here, thanks for the suggestion.

spec.md
pmengelbert: no reason -- I reverted this back to `<repository_name>`

spec.md
dmcgowan: Do you have an update to push with the updated language?

spec.md
```diff
@@ -304,6 +311,28 @@ Location: <blob-location>
...
+and `<other_namespace>` is the namespace from which the blob should be mounted. This step is usually taken in place of the
+previously-described `POST` request to `/v2/<name>/blobs/uploads/` <sup>[end-4a](#endpoints)</sup> (which is used to initiate an
+upload session).
+
+The response to a successful mount MUST be `201 Created`, and MUST contain the following header:
+```
+Location: <blob-location>
+```
+
+The digest contained in the `Location` header MAY be different from that of the blob that was mounted. As such, a client
```
dmcgowan, line 330: Where did this wording come from related to using the digest?

The previous language was here https://github.com/opencontainers/distribution-spec/blob/f67bc11ba3a083a9c62f8fa53ad14c5bcf2116af/spec.md#cross-repository-blob-mount and is more clear. A client should NEVER use a digest from the registry without verifying it matches. Normally the client will just ignore this, I don't think it is correct to say the client should use it and verifying is expensive. Most registries use sha256 anyway, the case where it would differ is if the registry was using a different hashing algorithm.

jonjohnsonjr: > Most registries use sha256 anyway, the case where it would differ is if the registry was using a different hashing algorithm.

It would be great to be more explicit about this in the spec. The [digest canonicalization section](https://github.com/opencontainers/distribution-spec/blob/f67bc11ba3a083a9c62f8fa53ad14c5bcf2116af/spec.md#content-digests) in the original spec seems mostly to do with signed schema 1 manifests:

> To provide verification of http content, any response MAY include a Docker-Content-Digest header. This will include the digest of the target entity returned in the response. For blobs, this is the entire blob content. For manifests, this is the manifest body without the signature content, also known as the JWS payload. Note that the commonly used canonicalization for digest calculation MAY be dependent on the mediatype of the content, such as with manifests.

Though the section about different algorithms is good and maybe we should copy that over:

> The client MAY choose to ignore the header or MAY verify it to ensure content integrity and transport security. This is most important when fetching by a digest. To ensure security, the content SHOULD be verified against the digest used to fetch the content. At times, the returned digest MAY differ from that used to initiate a request. Such digests are considered to be from different domains, meaning they have different values for algorithm. In such a case, the client MAY choose to verify the digests in both domains or ignore the server's digest. To maintain security, the client MUST always verify the content against the digest used to fetch the content.
> 
> IMPORTANT: If a digest is used to fetch content, the client SHOULD use the same digest used to fetch the content to verify it. The header Docker-Content-Digest SHOULD NOT be trusted over the "local" digest.

We already link out to https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#digests but that links to https://github.com/opencontainers/image-spec/blob/v1.0.1/canonicalization.md#canonicalization which is a 404 🤷 

Looks like it should be https://github.com/opencontainers/image-spec/blob/v1.0.1/considerations.md#canonicalization but I don't see anything here that would make me think "oh, I should make sure everything is using the same algorithm", especially in the context of a client-registry interaction

spec.md
pmengelbert: oops 🥵

spec.md
```diff
@@ -54,7 +58,7 @@ Several terms are used frequently in this document and warrant basic definitions
 - **Pull**: the act of downloading Blobs and Manifests from a Registry
 - **Blob**: the binary form of content that is stored by a Registry, addressable by a Digest
 - **Manifest**: a JSON document which defines an Artifact. Manifests are defined under the OCI Image Spec <sup>[apdx-2](#appendix)</sup>
-- **Config**: a section in the Manifest (and associated Blob) which contains Artifact metadata
+- **Config**: a blob referenced in the Manifest which contains Artifact metadata
```
jonjohnsonjr, line 61: This is also described in [image-spec](https://github.com/opencontainers/image-spec/blob/v1.0.1/config.md) if you want to link to that (similar to how manifest links to apdx-2).

pmengelbert: Thanks again, just pushed changes addressing this and the above

spec.md
pmengelbert: Just pushed a commit that reflects these suggestions.

dmcgowan: I agree, there may be a little bit of gap in specification around using sha256. There are a few spots where there is an assumption made and if the registry were to use a different algorithm, it could cause some issues. When starting an upload, the client should know the desired digest (but there are cases it doesn't). If it doesn't, and the client/registry are using different algorithms, the finalization would require the registry to either recalculate the digest, calculate it using multiple algorithms, or reject. In the case of cross repository mount, the digest is likely referenced in a manifest and it is not an option to use a different digest that the registry returns.

spec.md
```diff
@@ -304,6 +317,26 @@ Location: <blob-location>
...
+previously-described `POST` request to `/v2/<name>/blobs/uploads/` <sup>[end-4a](#endpoints)</sup> (which is used to initiate an
+upload session).
+
+The response to a successful mount MUST be `201 Created`, and MUST contain the following header:
+```
+Location: <blob-location>
+```
+
+
+Alternatively, if a registry does not support cross-repository mounting, it SHOULD return a `202`, indicating that the
```
jonjohnsonjr, line 337: > if a registry does not support cross-repository mounting

It's possible that the registry supports cross-repository mounting but is unable to mount the requested blob for various reasons (malformed request or blob non-existence, random distributed systems failures, etc.). May say something like "or is unable to mount the requested blob".

PR #208, opened 2020-10-29
Add test and spec section for HEAD requests

### LINKED ISSUE #207: Add HEAD endpoints
https://docs.docker.com/registry/spec/api/#existing-manifests

"Docker’s use of `HEAD` is also has a `Docker-Content-Digest` in the response headers so you can `HEAD` against a tag to determine the digest that the tag currently points to."

pmengelbert:
resolves #207

### DIFF
--- a/conformance/01_pull_test.go
+++ b/conformance/01_pull_test.go
@@ -81,6 +81,24 @@ var test01Pull = func() {
 		})
 
 		g.Context("Pull blobs", func() {
+			g.Specify("HEAD request to nonexistent blob should result in 404 response", func() {
+				SkipIfDisabled(pull)
+				req := client.NewRequest(reggie.HEAD, "/v2/<name>/blobs/<digest>",
+					reggie.WithDigest(dummyDigest))
+				resp, err := client.Do(req)
+				Expect(err).To(BeNil())
+				Expect(resp.StatusCode()).To(Equal(http.StatusNotFound))
+			})
+
+			g.Specify("HEAD request to existing blob should yield 200", func() {
+				SkipIfDisabled(pull)
+				req := client.NewRequest(reggie.HEAD, "/v2/<name>/blobs/<digest>",
+					reggie.WithDigest(configBlobDigest))
+				resp, err := client.Do(req)
+				Expect(err).To(BeNil())
+				Expect(resp.StatusCode()).To(Equal(http.StatusOK))
+			})
+
 			g.Specify("GET nonexistent blob should result in 404 response", func() {
 				SkipIfDisabled(pull)
 				req := client.NewRequest(reggie.GET, "/v2/<name>/blobs/<digest>",
@@ -100,6 +118,34 @@ var test01Pull = func() {
 		})
 
 		g.Context("Pull manifests", func() {
+			g.Specify("HEAD request to nonexistent manifest should return 404", func() {
+				SkipIfDisabled(pull)
+				req := client.NewRequest(reggie.HEAD, "/v2/<name>/manifests/<reference>",
+					reggie.WithReference(nonexistentManifest))
+				resp, err := client.Do(req)
+				Expect(err).To(BeNil())
+				Expect(resp.StatusCode()).To(Equal(http.StatusNotFound))
+			})
+
+			g.Specify("HEAD request to manifest path (digest) should yield 200 response", func() {
+				SkipIfDisabled(pull)
+				req := client.NewRequest(reggie.HEAD, "/v2/<name>/manifests/<digest>", reggie.WithDigest(manifestDigest)).
+					SetHeader("Accept", "application/vnd.oci.image.manifest.v1+json")
+				resp, err := client.Do(req)
+				Expect(err).To(BeNil())
+				Expect(resp.StatusCode()).To(Equal(http.StatusOK))
+			})
+
+			g.Specify("HEAD request to manifest path (tag) should yield 200 response", func() {
+				SkipIfDisabled(pull)
+				Expect(tag).ToNot(BeEmpty())
+				req := client.NewRequest(reggie.GET, "/v2/<name>/manifests/<reference>", reggie.WithReference(tag)).
+					SetHeader("Accept", "application/vnd.oci.image.manifest.v1+json")
+				resp, err := client.Do(req)
+				Expect(err).To(BeNil())
+				Expect(resp.StatusCode()).To(Equal(http.StatusOK))
+			})
+
 			g.Specify("GET nonexistent manifest should return 404", func() {
 				SkipIfDisabled(pull)
 				req := client.NewRequest(reggie.GET, "/v2/<name>/manifests/<reference>",
--- a/conformance/01_pull_test.go
+++ b/conformance/01_pull_test.go
@@ -97,6 +97,9 @@ var test01Pull = func() {
 				resp, err := client.Do(req)
 				Expect(err).To(BeNil())
 				Expect(resp.StatusCode()).To(Equal(http.StatusOK))
+				if h := resp.Header().Get("Docker-Content-Digest"); h != "" {
+					Expect(h).To(Equal(configBlobDigest))
+				}
 			})
 
 			g.Specify("GET nonexistent blob should result in 404 response", func() {
@@ -134,6 +137,9 @@ var test01Pull = func() {
 				resp, err := client.Do(req)
 				Expect(err).To(BeNil())
 				Expect(resp.StatusCode()).To(Equal(http.StatusOK))
+				if h := resp.Header().Get("Docker-Content-Digest"); h != "" {
+					Expect(h).To(Equal(manifestDigest))
+				}
 			})
 
 			g.Specify("HEAD request to manifest path (tag) should yield 200 response", func() {
@@ -144,6 +150,9 @@ var test01Pull = func() {
 				resp, err := client.Do(req)
 				Expect(err).To(BeNil())
 				Expect(resp.StatusCode()).To(Equal(http.StatusOK))
+				if h := resp.Header().Get("Docker-Content-Digest"); h != "" {
+					Expect(h).To(Equal(manifestDigest))
+				}
 			})
 
 			g.Specify("GET nonexistent manifest should return 404", func() {
--- a/spec.md
+++ b/spec.md
@@ -128,7 +128,7 @@ Typically, the first step in pulling an artifact is to retrieve the manifest. Ho
 
 ##### Pulling manifests
 
-To pull a manifest, perform a `GET` request to a url in the following form:
+To pull a manifest, perform a `GET` request to a URL in the following form:
 `/v2/<name>/manifests/<reference>` <sup>[end-3](#endpoints)</sup>
 
 `<name>` refers to the namespace of the repository. `<reference>` MUST be either (a) the digest of the manifest or (b) a tag name.
@@ -146,9 +146,9 @@ blob data.
 
 If the manifest is not found in the registry, the response code MUST be `404 Not Found`.
 
-##### Pulling Blobs
+##### Pulling blobs
 
-To pull a blob, perform a `GET` request to a url in the following form:
+To pull a blob, perform a `GET` request to a URL in the following form:
 `/v2/<name>/blobs/<digest>` <sup>[end-2](#endpoints)</sup>
 
 `<name>` is the namespace of the repository, and `<digest>` is the blob's digest.
@@ -156,6 +156,18 @@ To pull a blob, perform a `GET` request to a url in the following form:
 A GET request to an existing blob URL MUST provide the expected blob, with a response code that MUST be `200 OK`.
 
 If the blob is not found in the registry, the response code MUST be `404 Not Found`.
+
+##### Checking if content exists in the registry
+
+In order to verify that a repository contains a given manifest or blob, make a `HEAD` request to a URL in the following form:
+
+`/v2/<name>/manifests/<reference>` <sup>[end-12](#endpoints)</sup> (for manifests), or
+
+`/v2/<name>/blobs/<digest>` <sup>[end-11](#endpoints)</sup> (for blobs).
+
+A HEAD request to an existing blob or manifest URL MUST return `200 OK`.
+
+If the blob or manifest is not found in the registry, the response code MUST be `404 Not Found`.
 
 
 #### Push
@@ -467,6 +479,8 @@ of this specification.
 | end-9  | `DELETE` | `/v2/<name>/manifests/<reference>`                                | `202`                              | `404`/`400`/`405`               |
 | end-10 | `DELETE` | `/v2/<name>/blobs/<digest>`                                       | `202`                              | `404`/`405`                     |
 | end-11 | `POST`   | `/v2/<name>/blobs/uploads/?mount=<digest>&from=<other_namespace>` | `201`                              | `404`                           |
+| end-12 | `HEAD`   | `/v2/<name>/blobs/<digest>`                                       | `200`                              | `404`                           |
+| end-13 | `HEAD`   | `/v2/<name>/manifests/<reference>`                                | `200`                              | `404`                           |
 
 #### Error Codes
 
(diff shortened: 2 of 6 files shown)


pmengelbert:
@dmcgowan This is to address #207

markpeek:
Thank you for getting this PR put together. For HEAD on a manifest needs more information returned to be effective. In the docker implementation, Docker-Content-Digest is returned as a header which can then be used to compare against a cached copy of the manifest. I'd suggest adding a OCI-Content-Digest MUST be returned and Docker-Content-Digest MAY be returned (for interoperability). These likely should be added on a GET as well for consistency.

For a HEAD on a blob, it is debatable whether this is required given the digest is being used in the URL already. But the above headers could be added for consistency.

pmengelbert:
@markpeek Unfortunately, we cannot require Docker-Content-Digest. Those have been decided as "OPTIONAL", and replacing it with OCI-Content-Digest has not reached consensus yet. IIRC, the original spec specified that clients SHOULD NOT assemble the URL based on the contents of the `Docker-Content-Digest` header and instead SHOULD use what is provided in `Location`.

markpeek:
@pmengelbert Agreed on `Docker-Content-Digest` being optional which is why I said "MAY". Without a required digest being returned (such as via a OCI-Content-Digest header), the usefulness of HEAD for caching is diminished. If not in a header, how would a manifest pull with HEAD be able to verify the digest?

pmengelbert:
@markpeek manifests can be addressed by digest. I'm trying to make sure I understand the use-case,can you describe in a little more detail what is needed? Specifically, when verifying the digest in the `Docker-Content-Digest` header, is it being verified against a locally determined digest? Or is there some other digest against which it's being verified?

If you mean the former, the manifest can be addressed by digest, in which case a `HEAD` to `/v2/<name>/manifests/<digest>` would reveal whether or not the manifest with that digest exists, effectively verifying the local digest.

Again, feel free to correct me if I'm missing anything here!

markpeek:
I'm referring to a HEAD to `/v2/<name>/manifests/<reference>` where `reference` is a tag name such as `latest`.  Pulling this manifest likely will succeed but does not indicate whether it matches a cached copy or may need to be pulled to update. Having the digest provided allows for that check to be made without a full GET.

pmengelbert:
@markpeek I see, that makes sense. There will be further discussion necessary on this. We can bring it up at the next OCI call (Wednesday).  The likely solution is to have the `*-Content-Digest` be a SHOULD.

The issue with the header is messy, since we can't fully remove `Docker-...` since it may break existing compatibility, but we also don't want to encourage reliance on `Docker`-branded information. I agree with `OCI-Content-Digest` being a SHOULD, and perhaps `Docker-Content-Digest` can be a fallback (MAY). Thoughts? I think with this clarification we can reach consensus on the `OCI...` header soon.

Thanks for your help.

thaJeztah:
https://github.com/opencontainers/distribution-spec/pull/208#issuecomment-719640168

> These likely should be added on a GET as well for consistency.

To quote the HTTP specs [RFC2616, 9.4](https://tools.ietf.org/html/rfc2616#section-9.4)

```
The HEAD method is identical to GET except that the server MUST NOT
   return a message-body in the response. The metainformation contained
   in the HTTP headers in response to a HEAD request SHOULD be identical
   to the information sent in response to a GET request. This method can
   be used for obtaining metainformation about the entity implied by the
   request without transferring the entity-body itself. This method is
   often used for testing hypertext links for validity, accessibility,
   and recent modification.
```

So, if this header is added for `HEAD` requests, it _SHOULD_ also be send with `GET` requests (`HEAD` and `GET` being identical, except for the body being omitted)

The paragraph after that describes `HEAD` requests for purpose of caching, and is also somewhat relevant (although orthogonal to this proposal);

```
T he response to a HEAD request MAY be cacheable in the sense that the
   information contained in the response MAY be used to update a
   previously cached entity from that resource. If the new field values
   indicate that the cached entity differs from the current entity (as
   would be indicated by a change in Content-Length, Content-MD5, ETag
   or Last-Modified), then the cache MUST treat the cache entry as
   stale.
```

From the above; if a registry sends the relevant (Content-Length, Content-MD5, ETag or Last-Modified), clients could decide to skip a pull, and reuse the locally cached  manifest (and referenced data) even without the `OCI-Content-Digest` header

spec.md
```diff
@@ -145,6 +145,18 @@ A GET request to an existing blob URL MUST provide the expected blob, with a res
...
 
+##### Checking if content exists in the registry
+
+In order to verify that a repository contains a given manifest or blob, make a `HEAD` request to a URL in the following form:
+
+`/v2/<name>/manifests/<reference>` <sup>[end-12](#endpoints)</sup> (for manifests), or
+
+`/v2/<name>/blobs/<digest>` <sup>[end-11](#endpoints)</sup> (for blobs).
+
+A HEAD request to an existing blob or manifest URL MUST return `200 OK`.
```
thaJeztah, line 156: If this is meant to be returned as body, that would be in conflict with the HTTP specs (see the section I quoted); the body `MUST` be omitted

thaJeztah: (reading from my phone; I'll read up tomorrow when I'm back at my computer)

pmengelbert:
@markpeek I have adjusted the language in the spec based on what we discussed. If you are able to attend the OCI call (5pm US Eastern time on Wednesday), it would be helpful to have your more expert explanation of why this is needed. Thanks again for the tips.

@thaJeztah I have likewise changed the spec for `GET` requests for both manifests and blobs. Please voice your approval if all looks good.

Since these headers are OPTIONAL, no tests have been written to reflect the changes to the spec.

markpeek:
@pmengelbert thank you for making the changes. I do feel the `OCI-Content-Digest` header should be a 'MUST' to facilitate longer term interoperability. I will be on at least the first half of the call this week so we can discuss there with the rest of the team.

pmengelbert:
> @pmengelbert thank you for making the changes. I do feel the `OCI-Content-Digest` header should be a 'MUST' to facilitate longer term interoperability. I will be on at least the first half of the call this week so we can discuss there with the rest of the team.

@markpeek I have added you to the [agenda](https://hackmd.io/El8Dd2xrTlCaCG59ns5cwg#November-11-2020) for the call on Wednesday

jonjohnsonjr:
Looks like the next meeting is Dec 2, not this week?

jdolitsky:
@markpeek @jonjohnsonjr - just sent a message to the mailing list. I'm hoping we can still discuss this issue (and others) during that time slot if you are free, despite no official meeting

spec.md
```diff
@@ -124,27 +124,45 @@ Typically, the first step in pulling an artifact is to retrieve the manifest. Ho
...
+
+`/v2/<name>/manifests/<reference>` <sup>[end-12](#endpoints)</sup> (for manifests), or
+
+`/v2/<name>/blobs/<digest>` <sup>[end-11](#endpoints)</sup> (for blobs).
+
+A HEAD request to an existing blob or manifest URL MUST return `200 OK`. A successful response SHOULD contain the digest
+of the uploaded blob in the header `OCI-Content-Digest`. For legacy reasons, the digest MAY be contained in the header
+`Docker-Content-Digest` instead. Either header is OPTIONAL.
+
+If the blob or manifest is not found in the registry, the response code MUST be `404 Not Found`.
```
thaJeztah, line 165: As to supporting `HEAD` requests _in general_, I don't think registries `MUST` support `HEAD` requests. I would make that _optional_ (`SHOULD` or `MAY`). There's some reasons for that;

1. The `HEAD` requests are an _optimization_. Nothing should break if they're not supported, but clients may perform a pull where they could have used their local cache if they do not use `HEAD` requests, or if the registry doesn't support them.
2. `HEAD` requests may not be available in all situations (I've had users report that `HEAD` requests were blocked by proxies)
3. The spec should probably give some guidance on how clients should handle `HEAD` requests; the standard pattern is that clients _must_ (_should_ ?) fall back to `GET` requests if a `HEAD` request fails

_However_ `3.` is where a (big) challenge is; both to keep it _optional_, and to remain backward compatible with current implementations of registries.

The situation is that any current registry may not support `HEAD` requests, and because of that return a `404 Not found`. Because of that, `404` is not ambiguous, as it could either mean that;

a. the registry supports `HEAD` requests, but the given manifest was not found
b. the registry does not implement `HEAD` requests
c. `HEAD` requests are blocked (?)

This may be a problem, because now effectively clients `MUST` try the `GET` endpoint on any `404` status to know if the manifest exists (if the `GET` response also returns a `404`, the manifest does not exist).

Not sure what the best answer is  to distinguish `404 "not implemented"` from `404 "no such manifest"`. Perhaps the registry should return an empty `OCI-Manifest-Digest: ""` header to indicate _"I looked, but couldn't find"_; the header would not be present in situations `b.` and `c.`, so the client can fall back to doing a `GET` request in that case.

I should note that current versions of the registry (at least https://github.com/docker/distribution) does not send the header on `HEAD` requests if the manifest is not found, so the client would have to perform two requests for those;

```console
$ curl -X HEAD  -I \
   -H 'Accept: application/vnd.docker.distribution.manifest.v2+json' \
   -H 'Accept: application/vnd.docker.distribution.manifest.list.v2+json' \
   -H 'Accept: application/vnd.docker.distribution.manifest.v1+json' \
   http://localhost:5000/v2/nosuchrepo/manifests/latest

HTTP/1.1 404 Not Found
Content-Type: application/json
Docker-Distribution-Api-Version: registry/2.0
X-Content-Type-Options: nosniff
Date: Wed, 11 Nov 2020 21:12:11 GMT
Content-Length: 96
```

thaJeztah (changes requested):

spec.md
```diff
@@ -124,27 +124,44 @@ Typically, the first step in pulling an artifact is to retrieve the manifest. Ho
...
-To pull a manifest, perform a `GET` request to a url in the following form:
+To pull a manifest, perform a `GET` request to a URL in the following form:
 `/v2/<name>/manifests/<reference>` <sup>[end-3](#endpoints)</sup>
 
 `<name>` refers to the namespace of the repository. `<reference>` MUST be either (a) the digest of the manifest or (b) a tag name.
 The `<reference>` MUST NOT be in any other format.
 
 A GET request to an existing manifest URL MUST provide the expected manifest, with a response code that MUST be `200 OK`.
+A successful response SHOULD contain the digest of the uploaded blob in the header `OCI-Content-Digest`. For legacy
+reasons, the digest MAY be contained in the header `Docker-Content-Digest` instead. Either header is OPTIONAL.
```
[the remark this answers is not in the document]
jonjohnsonjr: Would it make sense to have a backwards compatibility section somewhere in the spec? It could include `Docker-Content-Digest` and https://docs.docker.com/registry/spec/manifest-v2-2/#backward-compatibility

Some precedent in the image-spec: https://github.com/opencontainers/image-spec/blob/master/media-types.md#compatibility-matrix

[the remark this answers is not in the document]
thaJeztah: Yes, I think that's a good suggestion. It's good to keep "legacy" and "specification" separate, but given that we're still somewhat in a transition, it's hard to _fully_ ignore the old standards (it's not just registries, but many older clients are still active in the wild). Having sections dedicated to guidelines for "how to get a smooth transition" / "things to consider for backward compatibility" seems like a nice middle ground.

thaJeztah:
> I do feel the OCI-Content-Digest header should be a 'MUST' to facilitate longer term interoperability. 

Looks like I overlooked this comment; @markpeek could you elaborate on why it should be a `MUST`?

I would be hesitant making it a requirement, as (AFAICS) it's effectively an _optimization_  (also see my inline comments); registries can function without this header being present (at the possible cost of performing a `GET` request and parsing the body). I realise for the "docker hub rate limiting" case, it's definitely _useful_ to have (and if this proposal gets accepted will be added in addition to the existing `Docker-` prefixed headers), but I don't see a clear reason to _enforce_ it.

Requiring these headers will also complicate hosting static registries, which (as far as I know) was a goal of the specification.

spec.md
```diff
@@ -467,6 +485,8 @@ of this specification.
 | end-9  | `DELETE` | `/v2/<name>/manifests/<reference>`                                | `202`                              | `404`/`400`/`405`               |
 | end-10 | `DELETE` | `/v2/<name>/blobs/<digest>`                                       | `202`                              | `404`/`405`                     |
 | end-11 | `POST`   | `/v2/<name>/blobs/uploads/?mount=<digest>&from=<other_namespace>` | `201`                              | `404`                           |
+| end-12 | `HEAD`   | `/v2/<name>/blobs/<digest>`                                       | `200`                              | `404`                           |
```
dmcgowan, line 488: Could the value under `Method` for these endpoints be `GET` / `HEAD` rather than treating them as a separate endpoint entry.

dmcgowan (changes requested):
I am still -1 to defining a new `OCI-Content-Digest` header for 1.0 since this is not in use by any client or registry today. Test changes look good.

thaJeztah:
> This may be a problem, because now effectively clients MUST try the GET endpoint on any 404 status to know if the manifest exists (if the GET response also returns a 404, the manifest does not exist).
>
> Not sure what the best answer is to distinguish 404 "not implemented" from 404 "no such manifest".

Small update on that comment I placed earlier. I was wrong there, as an _endpoint_ that's implemented but which _doesn't_ implement a method (e.g., only supports `GET`, but not `POST` or `HEAD`) should return a `405`, not a `404` (thanks @tianon for bringing that up in a discussion); [RFC2616, 10.4.6](https://tools.ietf.org/html/rfc2616#section-10.4.6):

```markdown
10.4.6 405 Method Not Allowed

   The method specified in the Request-Line is not allowed for the
   resource identified by the Request-URI. The response MUST include an
   Allow header containing a list of valid methods for the requested
   resource.
```

So it should still return a `Allow` header to specify what _is_ accepted.

That said, MDN has this addendum; (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#Client_error_responses)

> The two mandatory methods, `GET` and `HEAD`, must never be disabled and should not return this error code.

I yet have to dig where that's described in the HTTP specs.

spec.md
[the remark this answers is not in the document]
jdolitsky: Based on several conversations about this: a new OCI- header will not be introduced. A subsequent release may introduce this header if it becomes necessary

conformance/01_pull_test.go
```diff
@@ -100,6 +118,34 @@ var test01Pull = func() {
 		})
 
 		g.Context("Pull manifests", func() {
+			g.Specify("HEAD request to nonexistent manifest should return 404", func() {
```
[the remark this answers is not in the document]
pmengelbert: @thaJeztah 
1. This is covered in the HTTP RFC you mentioned, so it's out of the scope of these tests
2. We went with `Docker-Content-Digest`, and it's being measured against a real digest. It won't pass the test if a tag name is returned
3. See the new check we've added, which verifies that the digest of the bytes matches the digest returned by the header.

spec.md
pmengelbert: @thaJeztah For now, we're requiring that HEAD is supported (MUST), but if a manifest is not found the check for the `Docker-Content-Digest` header is not performed.

pmengelbert: @thaJeztah For now, we're requiring that HEAD is supported (MUST), but if a manifest is not found the check for the `Docker-Content-Digest` header is not performed.

spec.md
pmengelbert: @dmcgowan sure, just pushed a change to that effect, thanks.

pmengelbert:
@jdolitsky @dmcgowan Just did a round of updates, can I get another review? Thanks

pmengelbert:
Also added a small fix for #215

jdolitsky (approved):
LGTM

spec.md
jzelinskie: >The situation is that any current registry may not support HEAD requests, and because of that return a 404 Not found.

Technically, you should return 405 (Method not Allowed) instead of 404 if you don't support HEAD requests (or any other method for that matter). I know this ideal, but we could make that the differentiator for explicitly not supporting HEAD.

dmcgowan (approved):
LGTM

Issue #212, opened 2020-11-12
Proposal: add (optional) section on content-negotiation


thaJeztah:
The distribution spec currently does not describe active content negotiation, leaving it up to individual implementations whether (and _how_) content negotiation should be performed.

This proposal is still a draft, and many "blanks" to fill, but I thought I'd open it, to allow having a discussion on this topic.

## 1. "Problem" statement

Currently, the client is responsible for picking the right variant for multi-manifest ("multi arch") repositories. While this works well, it requires at multiple steps, and multiple requests to get the image manifest;

1. (`GET /v2`)
2. (optional) check if cache can be used: `HEAD /v2/<name>/manifests/<reference>`
3. fetch manifest-list: `GET /v2/<name>/manifests/<reference>`
4. (parse manifest-list and resolve digest for best match)
5. fetch image manifest: `GET /v2/<name>/manifests/<DIGEST>`
6. fetch blob: `/v2/<name>/blobs/<digest>` (repeat for each blob)

Content-negotiation can avoid multiple request, which could bring some performance improvements, and (if implemented by the registry), simplify logic in the client.

In addition, content-negotiation can assist in providing backward-compatibility (as outlined in the next section).

## 2. Existing uses of content-negotiation

Content negotiation is already in use by some registry implementations, such as Docker Hub (mostly for backward-compatibility). Here's some tests against docker hub;

> **Note**: for formatting on GitHub, I removed `application/vnd.docker.distribution` prefixes for the content-types in the table below.

| `Accept`                                                                    | Multi?             | Result                  | Notes                                                                                        |
|:----------------------------------------------------------------------------|:-------------------|:------------------------|:---------------------------------------------------------------------------------------------|
| `<not present>`                                                             | :white_check_mark: | `manifest.v1+prettyjws` | `v1`, `linux/amd64`, for backward compatibility with old, non-v2 clients                     |
| `*/*`                                                                       | :white_check_mark: | `manifest.v1+prettyjws` | `v1`, `linux/amd64`, for backward compatibility with old, non-v2 clients (same as above)     |
| `application/json`                                                          | :white_check_mark: | `manifest.v1+prettyjws` | `v1`, `linux/amd64`, for backward compatibility with old, non-v2 clients (same as above)     |
| `manifest.v1+json`                                                          | :white_check_mark: | `manifest.v1+prettyjws` | `v1`, matching `Accept` header (`linux/amd64` for backward compatibility)                    |
[...]

PR #218, opened 2020-12-09
Add placeholder for content negotiation

jdolitsky:
Related to #212

### DIFF
--- a/content-negotiation.md
+++ b/content-negotiation.md
@@ -1,0 +1,4 @@
+## Content Negotiation
+
+TODO - Please see
+[issue #212](https://github.com/opencontainers/distribution-spec/issues/212).
--- a/spec.md
+++ b/spec.md
@@ -122,6 +122,7 @@ In order to test a registry's conformance against these workflow categories, ple
 ### Workflow Categories
 
 #### Pull
+
 The process of pulling an artifact centers around retrieving two components: the manifest and one or more blobs.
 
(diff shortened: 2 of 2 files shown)


mikebrow (approved):
LGTM

dmcgowan (approved):
LGTM

jonjohnsonjr:
@justincormack what do you think about this? I would also like to drop this entirely in favor of manifest list based mechanisms, but I'm reluctant to drop it entirely until registries stop downconverting to schema 1.

justincormack (approved):

justincormack:
@jonjohnsonjr yeah thats fine. Note that the spec for Accept https://tools.ietf.org/html/rfc7231#section-5.3.2 says

  "If the header field is
   present in a request and none of the available representations for
   the response have a media type that is listed as acceptable, the
   origin server can either honor the header field by sending a 406 (Not
   Acceptable) response or disregard the header field by treating the
   response as if it is not subject to content negotiation."

So it is always fine to serve something the client did not ask for, and
totally ignore Accpet headers. I think we should recommend never sending a 406.

jonjohnsonjr:
> origin server can either honor the header field by sending a 406 (Not
> Acceptable) response or disregard the header field by treating the
> response as if it is not subject to content negotiation.

Ooh this is nice. Will probably update GCR's behavior here to match. We might want to call out specifically some suggested behavior in the placeholder section around this? Right now we are returning a 404, which felt correct at the time, but is obviously non-conforming per the rfc. Just ignoring it would be wonderful, but I'm a bit concerned about breaking clients.

Issue #234, opened 2021-02-04
Proposal: Deprecate single-request monolithic blob uploads before 1.0


jonjohnsonjr:
## background

It is possible to upload a blob in a single POST, per the spec, called a monolithic upload.

It is also possible to upload a blob via POST/PUT, which is also called a monolithic upload.

I'm talking about the first one. I call this "single request monolithic upload", which may be abbreviated as "SRMU" in the rest of this proposal.

I've put specific, actionable proposals in **bold**.

I can send a PR that implements my proposal, but I'd like to solicit some feedback first to see if anyone feels strongly opposed.

## proposal

1. **We should remove SRMUs from the spec.** As with catalog, the lack of broad support and high cost make it unfit for the spec.
2. **We should clarify the ambiguity between 201 and 202 response codes**, whether we deprecate it or not.

_(Optional aside (3): **We should have a deprecation document.** At one point we deprecated catalog, but I can't seem to find reference to it anywhere after https://github.com/opencontainers/distribution-spec/pull/178. It can be helpful to clearly document the differences, as in https://github.com/opencontainers/image-spec/pull/817)_

### ambiguity

This API is not well defined, so it is unclear how clients should interpret the registry's response.

For cross-repo mounting, if the mount failed for any reason, the registry will return a 202 that indicates you should proceed with the upload as usual. If the mount succeeds, the registry will return a 201, and clients can skip the upload.

For SRMUs, we don't have such a distinction... from the current spec at HEAD:

> Successful completion of the request MUST return either a 201 Created or a 202 Accepted

See https://github.com/opencontainers/distribution-spec/pull/230#discussion_r568762650, where a test enforcing the spec (202 is allowed) would pass for a registry even though it doesn't support it.

**At the very least, we should update the spec to say that clients should interpret 201 as "upload succeeded" and 202 as "proceed with upload at `Location`"**

This has the nice failure mode for registries that don't support SRMUs. If a registry doesn't bother to check the request body or `?digest` querystring parameter of POST requests, they will likely just return a 202 with a `Location`, as per a normal chunked upload. Clients that see this should probably assume that the registry does not support SRMU in this case, and proceed with a chunked upload or POST/PUT monolithic upload.

### lack of support

From https://github.com/opencontainers/distribution-spec/issues/211#issuecomment-726312919

[...]

PR #221, opened 2020-12-12
Adding faq again to replace #189

vsoch:
Third time is the charm! No I'm not crying, you're crying! :laughing: 

Signed-off-by: vsoch <vsochat@stanford.edu>

PR #239, opened 2021-03-03
Describe single-request monolithic upload behavior

jonjohnsonjr:
This addresses the uncontroversial part of https://github.com/opencontainers/distribution-spec/issues/234 by describing how clients should interpret the response codes from registries who do not support single request monolithic uploads.

I am happy to follow up with a deprecation doc, but the existing language from https://github.com/opencontainers/distribution-spec/pull/221 is actually pretty reasonable and expresses some of what I had intended to convey by deprecating this.

The PR diff is not great because I flipped the order in which monolithic blob uploads are introduced, but I've split that into a separate commit to make it easier to review.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -197,34 +197,8 @@ There are two ways to push blobs: chunked or monolithic.
 ##### Pushing a blob monolithically
 
 There are two ways to push a blob monolithically:
-1. A single `POST` request
-2. A `POST` request followed by a `PUT` request
-
----
-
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
@dmcgowan @jdolitsky PTAL

jdolitsky (approved):
looks great

jdolitsky:
so we will leave conformance test then, for now?

dmcgowan (approved):
LGTM

PR #246, opened 2021-03-09
Modify conformance tests for single-shot uploads

jdolitsky:
This is an attempt to implement this suggestion: https://github.com/opencontainers/distribution-spec/pull/230#issuecomment-790126525

Also updated Go module dependencies, Go version, golanglint-ci version, and removed some dead code


jzelinskie:
Should we change this to be optional if so many registries don't support it currently?
If nobody is supporting it now, I don't think it's desirable to force them to considering this is basically legacy functionality.

jdolitsky:
@jzelinskie good call. just added a `OCI_SKIP_SINGLE_SHOT_UPLOAD_TEST` env var for this

jonjohnsonjr:
> Should we change this to be optional if so many registries don't support it currently?

This is basically testing that the non-support does "the right thing".

If they don't support it, then we would expect a 202 from the POST, and doing a blob GET should result in a 404. That will pass.

If they _do_ support it, then we would expect a 201 from the POST, and doing a blob GET should result in a 200. That will also pass.

jdolitsky:
Yes, it does:

Screen Shot 2021-03-09 at 4 30 47 PM

The single-shot POST returns a 202

jdolitsky:
I retract my comments about this breaking distribution.

On further inspection, the same config blob is uploaded during pull test setup. If this is run with the following environment, things pass with docker/distribution (distribution/distribution):

```
export OCI_NAMESPACE=oci-test
export OCI_TEST_CONTENT_MANAGEMENT=0
export OCI_TEST_PUSH=1
export OCI_TEST_PULL=0
export OCI_ROOT_URL=http://localhost:5000
export OCI_TEST_CONTENT_DISCOVERY=0
```

So if anything, there is a bug in the tests which disallow running all 4 categories at once

dmcgowan:
@jdolitsky is the same true for the other registries mentioned? If we can't find a legitimate breakage, we should just remove the skip and get this update in.

jdolitsky:
@dmcgowan as far as I understand, Harbor (the other registry mentioned), uses distribution as backend. I can't imagine this will break them with that being the case. Also considering 230 was opened by one of the maintainers

Will go ahead and remove the skip ability

jonjohnsonjr:
I think we need to fix #247 before we can merge this, right?

jdolitsky:
@jonjohnsonjr probably best to do so - just opened #249 to address

jdolitsky:
After rebasing, here are the results running against docker/distribution (distribution/distribution):

build conformance binary:
```
make conformance-binary
```

create config file with delete enabled:
```
cat > dist-config.yml << EOF
version: 0.1
log:
  level: debug
storage:
    delete:
      enabled: true
    filesystem:
        rootdirectory: /tmp/registry
http:
    addr: :5000
    headers:
        X-Content-Type-Options: [nosniff]
EOF
```

start registry at localhost 5000:
```
docker run -v "$(pwd)/dist-config.yml":/etc/docker/registry/config.yml --rm -p 5000:5000 registry:2.7.1
```

set environment, all categories enabled:
```
export OCI_CROSSMOUNT_NAMESPACE=oci-test-cross
export OCI_NAMESPACE=oci-test
export OCI_TEST_CONTENT_MANAGEMENT=1
export OCI_TEST_PUSH=1
export OCI_TEST_PULL=1
export OCI_ROOT_URL=http://localhost:5000
export OCI_TEST_CONTENT_DISCOVERY=1
```

run conformance tests:
```
output/conformance.test
```

results:
```
Ran 59 of 62 Specs in 0.426 seconds
FAIL! -- 55 Passed | 4 Failed | 0 Pending | 3 Skipped
```

The 201 vs. 202 is now passing, but failing in other areas previously described here: https://github.com/distribution/distribution/issues/3203

conformance/02_push_test.go
```diff
@@ -71,15 +71,23 @@ var test02Push = func() {
...
 				req := client.NewRequest(reggie.GET, "/v2/<name>/blobs/<digest>", reggie.WithDigest(configs[1].Digest))
 				resp, err := client.Do(req)
 				Expect(err).To(BeNil())
-				Expect(resp.StatusCode()).To(Equal(http.StatusOK))
+				if lastResponse.StatusCode() == http.StatusAccepted {
+					Expect(resp.StatusCode()).To(Equal(http.StatusNotFound))
+				} else {
+					Expect(resp.StatusCode()).To(Equal(http.StatusOK))
+				}
 			})
```
jdolitsky, line 91: this is the heart of the change

conformance/02_push_test.go
```diff
@@ -71,15 +71,23 @@ var test02Push = func() {
...
 				SkipIfDisabled(push)
 				req := client.NewRequest(reggie.GET, "/v2/<name>/blobs/<digest>", reggie.WithDigest(configs[1].Digest))
 				resp, err := client.Do(req)
 				Expect(err).To(BeNil())
-				Expect(resp.StatusCode()).To(Equal(http.StatusOK))
+				if lastResponse.StatusCode() == http.StatusAccepted {
+					Expect(resp.StatusCode()).To(Equal(http.StatusNotFound))
+				} else {
+					Expect(resp.StatusCode()).To(Equal(http.StatusOK))
+				}
```
mikebrow, line 90: looking good should there be a loop waiting for success after the initial (immediate) not found case?

jdolitsky: I believe that it will continue to be 404 until the blob is upload via the URL in `Location` header. So basically the upload is considered incomplete if it did not return a `201`

@jonjohnsonjr @dmcgowan - please correct me if I'm mistaken

jonjohnsonjr: Yeah, we aren't testing the client, we're probing the registry for its behavior... so there won't be any success unless we write some code to take appropriate action based on the registry response.

We _could_ have another test that finalizes the upload with a PUT if lastResponse.StatusCode() == http.StatusNotFound, which would test that the registry is able to receive the payload after a "failed" monolithic upload, but it doesn't seem super necessary... I'd leave that up to Josh.

This looks fine to me.

jdolitsky: This is tested in the very next test here: https://github.com/opencontainers/distribution-spec/blob/bbb8a94b0bfd7ea93ef257d6326fb4a4962cab51/conformance/02_push_test.go#L94

I think less requests is better, unless there is some issue/danger in having an unused upload session created as part of running these tests.

jonjohnsonjr: Agreed, that looks good to me.

mikebrow (approved):
LGTM

dmcgowan (approved):

PR #249, opened 2021-03-10
Use unique blobs for each workflow category

### LINKED ISSUE #247: Use unique blobs/configs for each workflow category
Re-using a config blob for setup during the pull tests is causing failures in push tests when all categories are run at once.

For more info, please see https://github.com/opencontainers/distribution-spec/pull/246#issuecomment-794506909

### LINKED ISSUE #248: Conformance tests run against the same repo twice causes tag/digest mismatch
Bug introduced in #208 when adding tests for HEAD requests. In order to test HEAD for a manifest by tag, the repo is listed for an existing tag. If using a tag that was pushed previously (from a previous test run), the digests do not match

jdolitsky:
Resolves #247 
Resolves #248

### DIFF
--- a/conformance/01_pull_test.go
+++ b/conformance/01_pull_test.go
@@ -1,7 +1,6 @@
 package conformance
 
 import (
-	"fmt"
 	"net/http"
 	"os"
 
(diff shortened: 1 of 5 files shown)


conformance/01_pull_test.go
```diff
@@ -71,6 +70,13 @@ var test01Pull = func() {
 				req := client.NewRequest(reggie.GET, "/v2/<name>/tags/list")
 				resp, _ := client.Do(req)
 				tag = getTagNameFromResponse(resp)
+
+				// attempt to forcibly overwrite this tag with the unique manifest for this run
+				req = client.NewRequest(reggie.PUT, "/v2/<name>/manifests/<reference>",
+					reggie.WithReference(tag)).
+					SetHeader("Content-Type", "application/vnd.oci.image.manifest.v1+json").
+					SetBody(manifests[0].Content)
+				client.Do(req)
```
dmcgowan, line 79: CI doesn't like this being unchecked

jdolitsky: @dmcgowan - resolved

dmcgowan (approved):
LGTM

Issue #250, opened 2021-03-12
Proposal: Document ETag and Precondition Semantics


jonjohnsonjr:
These are just [part of HTTP](https://tools.ietf.org/html/rfc7232) already, so we don't _really_ need to define them (IMO), but it would be nice if clients could rely on these headers.

We discussed this on the dev call a few weeks ago, filing this as a placeholder for a coming pull request to clearly document the expect behavior so that registry implementations can start to follow it.

jonjohnsonjr:
@awakecoding might make sense to continue your [ETag discussion](https://github.com/opencontainers/image-spec/pull/826#issuecomment-805415911) here.

awakecoding:
@jonjohnsonjr @SteveLasker excellent, let's discuss it here! Here is my primary goal: find a way to keep browser Etag compatibility (opaque value) while also passing a non-opaque Etag-like digest-based value, making it possible to implement caching *without* the need for associative data attached to a specific file, especially for manifests.

By this I mean being able to correctly recover the Etag-like value associated with a cached manifest file on disk without some additional metadata to store the corresponding Etag. This is very similar to the blobs that are digest-based, except made usable for manifests, and would greatly simplify efficient manifest update checks without uselessly pulling the manifest contents.

The **Etag** standard HTTP header is **opaque** by definition. Browsers and HTTP caches implement this mechanism such that if you cache an HTTP response along with the ETag value, you can make the same query again including the ETag value, and avoid receiving the response contents if it hasn't changed. This mechanism is almost perfect aside from the fact that the Etag value is **opaque** by definition, meaning we can't just declare it to be digest-based, and therefore non-opaque. A proper ETag implementation doesn't try interpreting the ETag value, it just stores it as a tag alongside the cache data.

This being said, ETag values are often digest-based, even though you cannot rely on them being digest-based. Here are the ETag and Docker-Content-Digest headers from a response I captured in Fiddler between the ORAS CLI and Azure Container Registry:

```
Docker-Content-Digest: sha256:083aa17315658e6c368cd46d58dc9e257c230e29b359c5b92008fe476955e099
Etag: "sha256:083aa17315658e6c368cd46d58dc9e257c230e29b359c5b92008fe476955e099"
```

[...]

PR #251, opened 2021-03-13
Document Conditional Requests

jonjohnsonjr:
Here's a first stab at https://github.com/opencontainers/distribution-spec/issues/250

This isn't complete, but I wanted to get something together so we can start bikeshedding.

A couple other use cases we should add are deleting tags (untagging) and interactions with caches.


jonjohnsonjr:
@sargun had a comment about registry support for this. Many registries supply ETags today, but it's unclear if they already support conditional requests.

One simple idea would be defining an HTTP header such that registries can positively indicate support for conditional requests.

E.g. something dumb like:

`OCI-Distribution-RFC7232: yes`

preconditions.md
```diff
@@ -0,0 +1,75 @@
+## Conditional Requests
+
+The [Distribution Specification](spec.md) describes mutable tag identifiers, which can be used when pushing, pulling, and deleting manifests.
+Given that HTTP is a stateless protocol for distributed systems, the mutability of tag references leads to many possible race conditions.
+Fortunately, HTTP has a mechanism for dealing with race conditions: conditional requests.
+
+[RFC 7232](https://tools.ietf.org/html/rfc7232) defines the semantics of Conditional Requests in great detail, so this document will simply highlight the relevant portions of the RFC as applied to the Distribution Specification.
+
+### Safely Mutating Manifests
```
stevvooe, line 9: Is this limited to only mutating a manifest for a given tag?

jonjohnsonjr: I can't really think of anywhere else this is needed, but I might be overlooking something. There are races when creating, moving, and deleting tags, but everything else should be safe in the face of concurrency... right?

I think it's fine for this to apply to e.g. a manifest PUT, by digest, but it seems not very useful. You could say "don't PUT this manifest if it already exists", but PUT should be idempotent regardless, and you either want the manifest to exist or not.

sargun:
> @sargun had a comment about registry support for this. Many registries supply ETags today, but it's unclear if they already support conditional requests.
> 
> One simple idea would be defining an HTTP header such that registries can positively indicate support for conditional requests.
> 
> E.g. something dumb like:
> 
> OCI-Distribution-RFC7232: yes
> 
> If a client sees that header, they could rely on conditional request semantics. If they don't see that header, they could choose to alter their behavior to work around potential race conditions (if possible), log a warning, return an error, etc.

What API would the registry have to add this header for -- all? It seems like to me, the client can be optimistic, and if the registry doesn't support it, it can return some known error value.

jonjohnsonjr:
> What API would the registry have to add this header for -- all?

I would say anywhere that includes an `Etag` header? It would indicate that the client can rely on that `Etag` for conditional requests. Its absence would (likely) indicate a lack of support, so clients would have to proceed with caution.

> It seems like to me, the client can be optimistic, and if the registry doesn't support it, it can return some known error value.

I don't think that works. If a registry doesn't implement conditional requests, the registry would happily just allow any requests to go through and not return any errors for conflicts. We need the registry to signal positive support, because by default there is no way to determine the lack of support beyond probing the registry's behavior by sending a request you expect to fail.

preconditions.md
jonjohnsonjr: I guess we could also use this when uploading blobs?

Instead of always doing a HEAD request to check for blob existence in a registry, you could just start with a POST that includes the `If-None-Match` header. The registry returning a 412 would indicate that the blob already exists, and we could do both the existence check and the upload initiation in a single request.

Unfortunately, this would require us to have [deterministic Etags](https://github.com/opencontainers/distribution-spec/issues/250#issuecomment-806022857), as @awakecoding is suggesting, which I like, but I haven't thought completely through.

jdolitsky:
@jonjohnsonjr - can you address @vbatts and @jzelinskie comments above and we can merge?

Also - are these things we can realistically add to the conformance suite?

stevvooe (approved):
LGTM on the whole idea. Might need to work out a few nits.

imjasonh:
Friendly ping. Some of the recent discussion in the reference types WG has bumped into this as a blocker, and even regardless of how/whether that work uses and updates indexes, it would be great to have this formally spelled out.

@opencontainers/distribution-spec-maintainers any more comments and/or approvals?

cc @chris-crone @tonistiigi

tonistiigi:
>  If a registry doesn't implement conditional requests, the registry would happily just allow any requests to go through and not return any errors for conflicts.

Can we be sure that there isn't a case where the current registry only uses etags for `GET` and fails on all `If-Match` because it doesn't understand the condition(eg. `PUT/GET` are managed by different internal services)? I guess even in that case client could do a `GET` again, and see that the etag has not changed so the server must be misbehaving.

jdolitsky (approved):

Issue #52, opened 2019-01-31
Tags are essentially undefined in the spec


jonjohnsonjr:
The spec references "tags" quite a bit, but never defines them. I'm curious what everyone here (mostly registry operators) think a tag actually is.

My understanding of a tag is that it is effectively a git ref, i.e. it is a pointer to a manifest (or image index) digest that is namespaced to a repository.

I've been part of a conversation where users expected a repository to support having a tag for each platform so that they could push each platform's image individually and have the registry assemble a manifest list for them (cc @tianon please correct me if I'm wrong here).

Similarly, on [this thread](https://groups.google.com/a/opencontainers.org/d/msg/dev/oguystbnnw4/2f5Xe6a9DAAJ), some users expect the registry to support having a tag for each media type.

I wonder if it's worth clarifying somewhere (maybe not the spec, but just in this issue) so we can point to the answer of why we can't do this. It's clear that we can support pulling heterogeneous media types and platforms (that's the point of an image index), but we would require clients to assemble them rather than the registry (due to the trust model -- I expect the digest of what I pull to be what I pushed; you can't easily sign something the registry synthesizes).

cc @stevvooe @dmcgowan would love to hear your thoughts here.

PS: related, there is a `TAG_INVALID` error: `During a manifest upload, if the tag in the manifest does not match the uri tag, this error will be returned.`. cc @atlaskerr related to #47 this is a schema-1-ism.

mikebrow:
`A tag name must be valid ASCII and may contain lowercase and uppercase letters, digits, underscores, periods and dashes. A tag name may not start with a period or a dash and may contain a maximum of 128 characters.`

from: https://docs.docker.com/engine/reference/commandline/tag/

TagRegexp = match(`[\w][\w.-]{0,127}`)

mikebrow:
What is a tag? it's a string that passes the above regexp.
If a tag name is not provided but is required "latest" may be used as the default.

jonjohnsonjr:
So that's the format of the tag. Should we inline that into the spec? I see a regex for repositories but not tags.

I'm more interested in the semantics of a tag. It may be that we don't want to define how tags should work in order to grant registry operators flexibility in implementation, but right now it's apparently confusing for clients.

mikebrow:
> So that's the format of the tag. Should we inline that into the spec? I see a regex for repositories but not tags.

I think so.. let's see if anyone else chimes in.

[...]

PR #256, opened 2021-03-30
Define the representation for a tag name.

briandealwis:
A tag is defined in the spec as a manifest identifier, but the spec does not define the acceptable representations for a tag.  This PR provides a definition for the tag name based on the historical description from `docker tag`.

There was no reference to definitions for regular expressions, and other examples used basic regexs, so I didn't use character classes or bounds limits.

Motivated by #52

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -137,10 +137,14 @@ Typically, the first step in pulling an artifact is to retrieve the manifest. Ho
 To pull a manifest, perform a `GET` request to a URL in the following form:
 `/v2/<name>/manifests/<reference>` <sup>[end-3](#endpoints)</sup>
 
-`<name>` refers to the namespace of the repository. `<reference>` MUST be either (a) the digest of the manifest or (b) a tag name.
+`<name>` refers to the namespace of the repository. `<reference>` MUST be either (a) the digest of the manifest or (b) a tag.
 The `<reference>` MUST NOT be in any other format. Throughout this document, `<name>` MUST match the following regular expression:
 
 `[a-z0-9]+([._-][a-z0-9]+)*(/[a-z0-9]+([._-][a-z0-9]+)*)*`
+
+Throughout this document, `<reference>` as a tag MUST be at most 128 characters in length and MUST match the following regular expression:
+
+`[a-zA-Z0-9_]+[a-zA-Z0-9._-]*`
 
 The client SHOULD include an `Accept` header indicating which manifest content types it supports.
 In a successful response, the `Content-Type` header will indicate the type of the returned manifest.
--- a/spec.md
+++ b/spec.md
@@ -144,7 +144,7 @@ The `<reference>` MUST NOT be in any other format. Throughout this document, `<n
 
 Throughout this document, `<reference>` as a tag MUST be at most 128 characters in length and MUST match the following regular expression:
 
-`[a-zA-Z0-9_]+[a-zA-Z0-9._-]*`
+`[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}`
 
 The client SHOULD include an `Accept` header indicating which manifest content types it supports.
 In a successful response, the `Content-Type` header will indicate the type of the returned manifest.


spec.md
```diff
@@ -142,6 +142,10 @@ The `<reference>` MUST NOT be in any other format. Throughout this document, `<n
 
 `[a-z0-9]+([._-][a-z0-9]+)*(/[a-z0-9]+([._-][a-z0-9]+)*)*`
 
+Throughout this document, `<reference>` as a tag name MUST be at most 128 characters in length and MUST match the following regular expression:
```
mikebrow, line 145: I didn't see the term `tag name` before read right over it.

Please s/tag name/tag/ both here and above on line 140.

mikebrow (commented):
see comment regarding `tag name` as a term.

briandealwis:
Oops sorry: these comments ended up in my hidden in my personal email.

@jonjohnsonjr I'm a bit hesitant:
  - **character set:** should be brought in by reference to other standards documents such as HTTP (which defines that the URI and header block must be processed as octets encoded as a superset of US-ASCII).
  - **regex:** there needs to be a reference to a specific form of regexs as POSIX regular expressions don't have `\w`, and RFCs usually use Augmented Backus-Naur Form (ABNF) notation specified in [RFC2234](https://tools.ietf.org/html/rfc2234).  But your regexp showed me that my own regex was incorrect (ok for a tag to start with underscore).
  - **registries should support at least 128 characters:** I'm a bit confused as `128,inf` seems to imply that a registry must support tags that have a minimum of 128 characters?

jonjohnsonjr:
> seems to imply that a registry must support tags that have a minimum of 128 characters?

If registries don't support at least 128 characters in a tag, then they will be "broken" from the perspective of a client that expects to be able to use up to 128 characters.

It's basically Postel's Law: clients need to be conservative in what they do, and registries should be liberal in what they accept.

Is that a good idea? Who knows. The bare minimum, IMO, is that registries should support up to 128 characters. Given that clients are probably not prepared to deal with tags longer than 128 characters, maybe registries should also not support longer tags than that, but it seems a little bit limiting...

mikebrow (approved):
LGTM

jonjohnsonjr:
LGTM but I'm not a maintainer :)

waynr (approved):
Looks good to me since it answers a question I brought up in slack. Not sure if I'm a maintainer on this project though -- probably not.

spec.md
```diff
@@ -137,11 +137,15 @@ Typically, the first step in pulling an artifact is to retrieve the manifest. Ho
...
 
-`<name>` refers to the namespace of the repository. `<reference>` MUST be either (a) the digest of the manifest or (b) a tag name.
+`<name>` refers to the namespace of the repository. `<reference>` MUST be either (a) the digest of the manifest or (b) a tag.
 The `<reference>` MUST NOT be in any other format. Throughout this document, `<name>` MUST match the following regular expression:
 
 `[a-z0-9]+([._-][a-z0-9]+)*(/[a-z0-9]+([._-][a-z0-9]+)*)*`
 
+Throughout this document, `<reference>` as a tag MUST be at most 128 characters in length and MUST match the following regular expression:
+
+`[a-zA-Z0-9_]+[a-zA-Z0-9._-]*`
```
joaodrp, line 147:
```suggestion
`[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}`
```

Is there a reason for not adding the quantifier?

For reference, `[\w][\w.-]{0,127}` is what we're using in [distribution/distribution](https://github.com/distribution/distribution). I'm fine with not using `\w`, but I'd keep the length.

joaodrp (approved):

sudo-bmitch (approved):

vbatts:
@briandealwis your commit are missing DCO signature

briandealwis:
Added a sign-off.

vbatts (approved):
🐙

briandealwis:
> Approval required by 1 of: @caniszczyk, @crosbymichael, @dmcgowan, @jdolitsky, @jzelinskie, @mikebrow, @stevvooe

jzelinskie (approved):
Looks really good in terms of keeping things surgical.

PR #286, opened 2021-06-23
Add CODEOWNERS

jzelinskie:
This will allow us to use GitHub to enforce approvals rather than an outside CI system like pullapprove.


jdolitsky (approved):

vbatts (approved):
even tho the OP has a wrong PR subject, we can roll with these changes.
I'll enable the CODEOWNERs for main branch.

Issue #281, opened 2021-06-22
Create Conformance test ensuring mount fails "open"


sargun:
As discussed in: https://github.com/opencontainers/distribution-spec/pull/275

In order to allow for blobs to be mounted using automatic mount discovery "smoothly", registries need to conform the the behaviour of starting an upload when cases such as the `from` parameter are missing from the URL during a mount.

sargun:
see: https://github.com/opencontainers/distribution-spec/pull/296

PR #296, opened 2021-09-15
Add conformance test verifying that cross-repo mount fails "open"

sargun:
In https://github.com/opencontainers/distribution-spec/issues/281, we stated
that cross-repo mounts without from should start an upload. This is a
requirement to make it so that clients can adopt the changes in
https://github.com/opencontainers/distribution-spec/pull/275 smoothly.

### DIFF
--- a/conformance/02_push_test.go
+++ b/conformance/02_push_test.go
@@ -204,6 +204,17 @@ var test02Push = func() {
 		})
 
 		g.Context("Cross-Repository Blob Mount", func() {
+			g.Specify("Cross-mounting of a blob without the from argument should yield session id", func() {
+				SkipIfDisabled(push)
+				req := client.NewRequest(reggie.POST, "/v2/<name>/blobs/uploads/",
+					reggie.WithName(crossmountNamespace)).
+					SetQueryParam("mount", dummyDigest)
+				resp, err := client.Do(req)
+				Expect(err).To(BeNil())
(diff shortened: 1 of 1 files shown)


conformance/02_push_test.go
```diff
@@ -204,6 +204,17 @@ var test02Push = func() {
...
 		g.Context("Cross-Repository Blob Mount", func() {
+			g.Specify("Cross-mounting of a blob without the from argument should yield session id", func() {
+				SkipIfDisabled(push)
+				req := client.NewRequest(reggie.POST, "/v2/<name>/blobs/uploads/",
+					reggie.WithName(crossmountNamespace)).
+					SetQueryParam("mount", dummyDigest)
+				resp, err := client.Do(req)
+				Expect(err).To(BeNil())
+				Expect(resp.StatusCode()).To(Equal(http.StatusAccepted))
+				Expect(resp.GetRelativeLocation()).To(ContainSubstring(crossmountNamespace))
```
jonjohnsonjr, line 215: I think this is not strictly necessary (though I see it in the next case as well :/). The location header may point to an external service which might not include the namespace in the URL. I think we mostly just want to ensure Location is non-empty.

jonjohnsonjr (approved):

jonjohnsonjr (approved):

jdolitsky (approved):

Issue #282, opened 2021-06-22
Create conformance test for automatic mount origin discovery


sargun:
See: https://github.com/opencontainers/distribution-spec/pull/275

We need to add a test for this behaviour to the conformance tests when registries have security "disabled".

PR #297, opened 2021-09-16
Add test for automatic cross mounting behaviour

sargun:
This adds conformance tests for automatic cross mounting. Usually, the test
should not run as it can give a false positive. The OCI_AUTOMATIC_CROSSMOUNT
must be set based on how the registry is expected to be configured.

### DIFF
--- a/conformance/setup.go
+++ b/conformance/setup.go
@@ -71,6 +71,7 @@ const (
 	envVarAuthScope                 = "OCI_AUTH_SCOPE"
 	envVarDeleteManifestBeforeBlobs = "OCI_DELETE_MANIFEST_BEFORE_BLOBS"
 	envVarCrossmountNamespace       = "OCI_CROSSMOUNT_NAMESPACE"
+	envVarAutomaticCrossmount       = "OCI_AUTOMATIC_CROSSMOUNT"
 
 	emptyLayerTestTag = "emptylayer"
 	testTagName       = "tagtest0"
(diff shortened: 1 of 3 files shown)


sargun:
"Fixes": https://github.com/opencontainers/distribution-spec/issues/282

conformance/setup.go
```diff
@@ -320,6 +323,9 @@ func init() {
 
 	skipEmptyLayerTest, _ = strconv.ParseBool(os.Getenv(envVarPushEmptyLayer))
 	deleteManifestBeforeBlobs, _ = strconv.ParseBool(os.Getenv(envVarDeleteManifestBeforeBlobs))
+	automaticCrossmountVal := ""
+	automaticCrossmountVal, runAutomaticCrossmountTest = os.LookupEnv(envVarAutomaticCrossmount)
+	automaticCrossmountEnabled, _ = strconv.ParseBool(automaticCrossmountVal)
```
jdolitsky, line 328: So if I understand correctly, the 201 test will only run if both `runAutomaticCrossmountTest` and `automaticCrossmountEnabled` are true?

and the 202 test if `runAutomaticCrossmountTest` is true and `automaticCrossmountEnabled` is false?

jdolitsky (approved):

conformance/setup.go
sargun: Correct.

jonjohnsonjr (approved):

vbatts (approved):

PR #311, opened 2021-12-01
Add handling of unknown fields to conformance tests

imjasonh:
This change also removes the direct dependency on Go types defined in
the image-spec repo, since this made it difficult to define new fields.

Instead, the types are copied into this repo and modified to remove
unused fields, and add a new unspecified field for use in the test.

Signed-off-by: Jason Hall <jasonhall@redhat.com>

---

Note: I haven't run this against any real registries yet, I wanted to get feedback about the general direction before proceeding too much further.

cc @jdolitsky our beloved conformance czar 👑

### DIFF
--- a/conformance/go.mod
+++ b/conformance/go.mod
@@ -8,7 +8,6 @@ require (
 	github.com/onsi/ginkgo v1.16.4
 	github.com/onsi/gomega v1.16.0
 	github.com/opencontainers/go-digest v1.0.0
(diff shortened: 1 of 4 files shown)


jdolitsky (approved):
Fine with the general direction. Also tested this against bundlebar and zot with no issues

jonjohnsonjr (approved):

PR #318, opened 2022-03-03
Add conformance test coverage for data field

imjasonh:
The `data` proposal was merged: https://github.com/opencontainers/image-spec/pull/826 🎉 

This adds some conformance test coverage that the new field is round-trippable through registries, helpfully piggybacking off of #311. Since registries don't validate that the contents match it doesn't seem fair to add that case to conformance, but it's something to consider in the future.

@jdolitsky

### DIFF
--- a/conformance/image.go
+++ b/conformance/image.go
@@ -35,6 +35,9 @@ type Descriptor struct {
 	// Size specifies the size in bytes of the blob.
 	Size int64 `json:"size"`
 
+	// Data specifies the data of the object described by the descriptor.
+	Data []byte `json:"data"`
+
 	// NewUnspecifiedField is not covered by image-spec.
 	// Registry implementations should still successfully store and serve
 	// manifests containing this data.
--- a/conformance/setup.go
+++ b/conformance/setup.go
@@ -228,7 +228,8 @@ func init() {
 				MediaType:           "application/vnd.oci.image.config.v1+json",
 				Digest:              godigest.Digest(configs[i].Digest),
 				Size:                int64(len(configs[i].Content)),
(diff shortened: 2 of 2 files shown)


jdolitsky:
fire. let me test this out and report back

jdolitsky (approved):
Tried on Bundle Bar, distribution/distribution, and zot. No failures related to the addition of this field

conformance/image.go
```diff
@@ -35,6 +35,9 @@ type Descriptor struct {
 	// Size specifies the size in bytes of the blob.
 	Size int64 `json:"size"`
 
+	// Data specifies the data of the object described by the descriptor.
+	Data []byte `json:"data"`
```
mikebrow, line 39: from the referenced commit 
```
	// Data is an embedding of the targeted content. This is encoded as a base64
 	// string when marshalled to JSON (automatically, by encoding/json). If
 	// present, Data can be used directly to avoid fetching the targeted content.
 	Data []byte `json:"data,omitempty"`
```

don't see how Data could be empty from this test bucket..  just a nit...

mikebrow (commented):
see nit

conformance/image.go
imjasonh: It doesn't even seem to matter either way. An empty `[]byte` with and without the struct tag results in `{}` in my tests in the Go Playground.

SteveLasker:
Adding #321 for tracking

jdolitsky:
@opencontainers/distribution-spec-maintainers - can we get another +1?

imjasonh:
Friendly ping.

vbatts (approved):

Issue #343, opened 2022-09-13
CI broken due to git-validation error


jdolitsky:
Output:
```
go get -u github.com/vbatts/git-validation
...
# golang.org/x/sys/unix
Error: ../../../go/pkg/mod/golang.org/x/sys@v0.0.0-[20](https://github.com/opencontainers/distribution-spec/actions/runs/3041892471/jobs/4911416912#step:4:21)220913175220-63ea55921009/unix/syscall.go:83:16: undefined: unsafe.Slice
Error: ../../../go/pkg/mod/golang.org/x/sys@v0.0.0-20220913175220-63ea559[21](https://github.com/opencontainers/distribution-spec/actions/runs/3041892471/jobs/4911416912#step:4:22)009/unix/syscall_unix.go:118:7: undefined: unsafe.Slice
Error: ../../../go/pkg/mod/golang.org/x/sys@v0.0.0-20[22](https://github.com/opencontainers/distribution-spec/actions/runs/3041892471/jobs/4911416912#step:4:23)0913175220-63ea55921009/unix/sysvshm_unix.go:33:7: undefined: unsafe.Slice
note: module requires Go 1.17
Error: make: *** [Makefile:85: .install.gitvalidation] Error 2
Error: Process completed with exit code 2.
```

cc @vbatts

vbatts:
thanks @sudo-bmitch for the quick solution

PR #345, opened 2022-09-13
Fix: #343 CI for gitvalidation

sajayantony:
Signed-off-by: Sajay Antony <sajaya@microsoft.com>

This only attempts at unblocking the CI for the repo at this point. We should have an issue to bump up the modules. Takes @sudo-bmitch's fix in part here - https://github.com/opencontainers/image-spec/blob/decdb1500673e7ef4fd912e0741f67a49e35d502/Makefile#L119-L120

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ header.html: .tool/genheader.go specs-go/version.go
 install.tools: .install.gitvalidation
 
 .install.gitvalidation:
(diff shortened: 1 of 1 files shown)


jdolitsky (approved):

sudo-bmitch (approved):
LGTM

PR #379, opened 2023-02-02
Specify subject-processed response header

### LINKED ISSUE #377: Add "subject processed" response header
One question we've circled around is how a client can know that a manifest it pushed containing a `subject` field was processed by the registry, to know whether the client should push an index to the subject's fallback tag.

One solution would be for the registry to respond to a POST of a manifest with `subject` with a response header indicating that it's 1.1-compliant and is going to update the referrers response for you.

Something like `X-OCI-Subject-Processed` -- better names welcome.

Related: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#unavailable-referrers-api

imjasonh:
Fixes #377

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -454,9 +454,9 @@ Client and registry implementations SHOULD expect to be able to support manifest
 
 ###### Pushing Manifests with Subject
 
-When processing a request for an image or artifact manfiest with the `subject` field, a registry implementation that supports the [referrers API](#listing-referrers) MUST respond with the response header `OCI-Subject-Processed: true` to indicate to the client that the registry understands the request's `subject`.
+When processing a request for an image or artifact manfiest with the `subject` field, a registry implementation that supports the [referrers API](#listing-referrers) MUST respond with the response header `OCI-Subject: <subject digest>` to indicate to the client that the registry processed the request's `subject`.
 
-When pushing an image or artifact manifest with the `subject` field and the [referrers API](#listing-referrers) returns a 404 or the `OCI-Subject-Processed: true` header was not set, the client MUST:
+When pushing an image or artifact manifest with the `subject` field and the [referrers API](#listing-referrers) returns a 404 or the `OCI-Subject` header was not set, the client MUST:
 
 1. Pull the current referrers list using the [referrers tag schema](#referrers-tag-schema).
 1. If that pull returns a manifest other than the expected image index, the client SHOULD report a failure and skip the remaining steps.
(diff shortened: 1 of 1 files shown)


jdolitsky (dismissed):

spec.md
```diff
@@ -454,7 +454,9 @@ Client and registry implementations SHOULD expect to be able to support manifest
 
 ###### Pushing Manifests with Subject
 
-When pushing an image or artifact manifest with the `subject` field and the [referrers API](#listing-referrers) returns a 404, the client MUST:
+When processing a request for an image or artifact manfiest with the `subject` field, a registry implementation that supports the [referrers API](#listing-referrers) MUST respond with the response header `OCI-Subject-Processed: true` to indicate to the client that the registry understands the request's `subject`.
```
jonjohnsonjr, line 457: I was thinking more along the lines of:

```
OCI-Subject: sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270
```

So the registry actually has to process the subject field and extract the digest to generate the response, which guarantees that the registry _did_ actually understand the `subject` field enough to generate this response, which gives us some assurance that it did something meaningful with it.

imjasonh: That's a good idea.

So just as a clarifying example, if this manifest gets pushed:

```
POST /v2/blah/blah/manifests/sha256:d54bcc13e190301b388a828178c02e3c3faa6c04a3a54bf90802c17ae892cebc
{
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "layers": [
    ...
  ],
  "subject": {
    "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270"
    ...
  }
}
```

(i.e., manifest `d54b` refers to manifest `5b0b`)

...then the response header should be `OCI-Subject: sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270`

This means the registry had to at least parse that JSON and extract the `subject` to be able to return it to you.

I like this.

What should the client do if the `OCI-Subject` response is present but not the subject it sent (`OCI-Subject: sha256:0000...`) -- fail probably? I don't think we need to add spec language for this, just curious mostly.

jonjohnsonjr: I think probably fail or use the fallback tag.

jonjohnsonjr (approved):

mikebrow (approved):
LGTM

PR #380, opened 2023-02-02
Specify filters-applied response header

### LINKED ISSUE #376: Move filtersApplied annotation to a response header
> If filtering is requested and applied, the response MUST include an annotation (org.opencontainers.referrers.filtersApplied) denoting that an artifactType filter was applied. If multiple filters are applied, the annotation MUST contain a comma separated list of applied filters.

I think this would make more sense to be a response header, e.g., `X-OCI-Referrers-Filters-Applied` (naming obviously open to debate).

A few reasons, in decreasing order of strength:

1. requesting a filter is done at the HTTP level, in a query parameter. An indication that the filter was respected should belong at the HTTP level, and a response header is a good place.
2. adding the annotation changes the index manifest contents, which feels kinda gross. If my filter doesn't change the result (I only have SBOMs and I filtered to only show SBOMs) then the index manifest response would be the same, except one includes that annotation.
3. It's a reserved annotation, but folks can still add that annotation to their index manifests and POST them. They might not even mean to; they might pull a referrers index manifest from `/referrers?artifactType=foo`, including the annotation, and push that unchanged to a registry that doesn't support referrers, to the fallback tag, with the annotation included. This might be fine, but gets confusing, and it's completely avoidable.

imjasonh:
Fixes #376

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -569,7 +569,7 @@ The `Link` header MUST be set according to [RFC5988](https://www.rfc-editor.org/
 The registry SHOULD support filtering on `artifactType`.
 To fetch the list of referrers with a filter, perform a `GET` request to a path in the following format: `/v2/<name>/referrers/<digest>?artifactType=<mediaType>` <sup>[end-12b](#endpoints)</sup>.
 If filtering is requested and applied, the response MUST include a header `OCI-Filters-Applied: artifactType` denoting that an `artifactType` filter was applied.
-If multiple filters are applied, the annotation MUST contain a comma separated list of applied filters.
+If multiple filters are applied, the header MUST contain a comma separated list of applied filters.
 
 Example request with filtering:
 
--- a/spec.md
+++ b/spec.md
@@ -568,7 +568,7 @@ The `Link` header MUST be set according to [RFC5988](https://www.rfc-editor.org/
 
 The registry SHOULD support filtering on `artifactType`.
 To fetch the list of referrers with a filter, perform a `GET` request to a path in the following format: `/v2/<name>/referrers/<digest>?artifactType=<mediaType>` <sup>[end-12b](#endpoints)</sup>.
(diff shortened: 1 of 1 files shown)


jdolitsky (approved):

spec.md
```diff
@@ -580,6 +580,7 @@ GET /v2/<name>/referrers/<digest>?artifactType=application/vnd.example.sbom.v1
 Example response with filtering:
 
 ```json
+OCI-Filters-Applied: artifactType
```
sajayantony, line 583: Should this mirror the annotation and be `OCI-Referrers-Filters-Applied`

jonjohnsonjr: I don't think we need that much language. For the annotation, it made sense, because you have no context about where it's used when it's just an annotation. In this case it's obvious that the header is relevant to the response to the referrers API by virtue of it being a header on the response to the referrers API.

sajayantony: Agreed and LGTM then.

spec.md
```diff
@@ -568,7 +568,7 @@ The `Link` header MUST be set according to [RFC5988](https://www.rfc-editor.org/
 
 The registry SHOULD support filtering on `artifactType`.
 To fetch the list of referrers with a filter, perform a `GET` request to a path in the following format: `/v2/<name>/referrers/<digest>?artifactType=<mediaType>` <sup>[end-12b](#endpoints)</sup>.
-If filtering is requested and applied, the response MUST include an annotation (`org.opencontainers.referrers.filtersApplied`) denoting that an `artifactType` filter was applied.
+If filtering is requested and applied, the response MUST include a header `OCI-Filters-Applied: artifactType` denoting that an `artifactType` filter was applied.
 If multiple filters are applied, the annotation MUST contain a comma separated list of applied filters.
```
jonjohnsonjr, line 572:
```suggestion
If multiple filters are applied, the header MUST contain a comma separated list of applied filters.
```

mikebrow (approved):
LGTM

PR #403, opened 2023-04-19
tag v1.1.0-rc.2

vbatts:
## to tag

this is would mean tagging 49af349 as `v1.1.0-rc.2` (now with more SemVer)

## Changes

1880a43 modify range check to not need bytes prefix
72a5455 Use artifactType from image manifest
b077648 Allow missing bytes= prefix
6520ed9 remove bytes= prefix on range header
18fdcbd fail CI if conformance fails
a5582e3 expect units in range header
f0122f6 move location check, allow missing bytes= prefix
bea9b6f Automate conformance testing on PR using zot
c6400af Upgrade to reggie v0.6.1 to fix issue with Accept header
fa4a69a skip TLS verification during conformance tests
27ae6c2 conformance: make blob mount and upload checks more strict
fc2e963 whitespace lol
a1d7ad1 Specify Warning header usage
851ea52 Add OCI-Chunk-Min-Length header
b962ef8 Update Jon Johnson's email
ab44270 Updating go modules in conformance
83a61e7 Bump golang.org/x/net in /conformance
6c52520 Bump golang.org/x/text from 0.3.6 to 0.3.8 in /conformance
ab32a6f spec and README: reduce to just image-manifest
e66aadc Update spec.md
5025143 OCI-Subject: <subject digest>
4e77264 Add Go 1.20 support
16cc054 Specify subject-processed response header
d06de52 Specify filters-applied response header
af2e081 Fix broken link to image indices in spec.
a330837 Add a patch status to recover failed requests
315c7e0 conformance.test01Pull: Check client errors before using response object
bd87eee docs: Update release process docs with checklist
95bed23 version: switch back to -dev

### DIFF
--- a/specs-go/version.go
+++ b/specs-go/version.go
@@ -25,7 +25,7 @@ const (
 	VersionPatch = 0
 
 	// VersionDev indicates development branch. Releases will be empty string.
-	VersionDev = "-rc.2"
+	VersionDev = "-rc.2-dev"
 )
 
 // Version is the specification version that the package types support.
--- a/specs-go/version.go
+++ b/specs-go/version.go
@@ -25,7 +25,7 @@ const (
 	VersionPatch = 0
 
 	// VersionDev indicates development branch. Releases will be empty string.
-	VersionDev = "-dev"
+	VersionDev = "-rc.2"
 )
 
 // Version is the specification version that the package types support.


specs-go/version.go
```diff
@@ -25,7 +25,7 @@ const (
 	VersionPatch = 0
 
 	// VersionDev indicates development branch. Releases will be empty string.
-	VersionDev = "-dev"
+	VersionDev = "-rc.2-dev"
```
jdolitsky, line 28: should be just `-rc2`:
```suggestion
	VersionDev = "-rc2"
```

jdolitsky: or `-rc.2` (no dev). Right?

jdolitsky:
See also https://github.com/opencontainers/image-spec/pull/1049

specs-go/version.go
vbatts: well, the idea is that `main` always stays with the `-dev` suffix, and we tag the commit in the middle

jdolitsky: right but why `"-rc.2-dev"` and not `"-rc.2"` ?

jdolitsky: going off of https://github.com/opencontainers/image-spec/pull/1049

jzelinskie (approved):

specs-go/version.go
jdolitsky: er, let me rephrase - why `"-rc.2-dev"` and not `"-dev"` ? (for the second commit)

jdolitsky (approved):

specs-go/version.go
vbatts: yeah ok fair. I can change it to just `-dev`

dmcgowan (approved):

sudo-bmitch (approved):
I updated the change list to include #401. It is in the git history for this, so it looks like it was just a race condition in writing the change list.

stevvooe (approved):

mikebrow (approved):
LGTM

Issue #416, opened 2023-05-22
conformance: make it easier to isolate a single test failure


rogpeppe:
The conformance tests are not all independent of one another. That is, when there's a failure, isolating that test (for example by changing `Specify` to `FSpecify`) and re-running the test suite will not yield the same test failure, because the tests have interdependencies.

To take one example, I am currently seeing this failure:

```
/home/rogpeppe/other/oci-distribution-spec/conformance/00_conformance_suite_test.go:12
  Push
  /home/rogpeppe/other/oci-distribution-spec/conformance/02_push_test.go:15
    Blob Upload Chunked
    /home/rogpeppe/other/oci-distribution-spec/conformance/02_push_test.go:152
      Get on stale blob upload should return 204 with a range and location [It]
      /home/rogpeppe/other/oci-distribution-spec/conformance/02_push_test.go:215

      Expected
          <int>: 405
      to equal
          <int>: 204

      /home/rogpeppe/other/oci-distribution-spec/conformance/02_push_test.go:220
```

The location of the failure is [here](https://github.com/opencontainers/distribution-spec/blob/7fcdf80f3c8b68e296bad0c5cd19aefe3f7265b1/conformance/02_push_test.go#L220).

If I try to run that test on its own to see exactly what HTTP requests are being sent and received, so to try to understand why this failure is happening and how to fix it, I'd expect to be able to focus on the test by changing line 215 to use `FSpecify`.

However, that doesn't work because on line 217, the test [refers to](https://github.com/opencontainers/distribution-spec/blob/7fcdf80f3c8b68e296bad0c5cd19aefe3f7265b1/conformance/02_push_test.go#L217) `prevResponse`, a variable outside of that test, that has been populated by [this test](https://github.com/opencontainers/distribution-spec/blob/7fcdf80f3c8b68e296bad0c5cd19aefe3f7265b1/conformance/02_push_test.go#L190).

This pattern is common throughout the conformance tests.

In general, it's a good idea for tests to be self-contained. Failing that, it would be nice if tests could at least document their dependencies so running a test on its own doesn't require error-prone tracing of dependencies.

PR #420, opened 2023-05-25
conformance: check response status before checking location

rogpeppe:
I noticed that when these tests are running against a server that returns the wrong status code (an error, for example), the failure that we see doesn't mention the status code, but only that the returned location is empty.

This change checks the status code first, as if that's wrong, the location is very unlikely to be relevant.

### DIFF
--- a/conformance/02_push_test.go
+++ b/conformance/02_push_test.go
@@ -233,8 +233,8 @@ var test02Push = func() {
 				resp, err := client.Do(req)
 				Expect(err).To(BeNil())
 				location := resp.Header().Get("Location")
-				Expect(location).ToNot(BeEmpty())
 				Expect(resp.StatusCode()).To(Equal(http.StatusAccepted))
+				Expect(location).ToNot(BeEmpty())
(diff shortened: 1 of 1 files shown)


sudo-bmitch (approved):

conformance/02_push_test.go
```diff
@@ -233,8 +233,8 @@ var test02Push = func() {
 				resp, err := client.Do(req)
 				Expect(err).To(BeNil())
 				location := resp.Header().Get("Location")
```
rchincha, line 235: nit: why not colocate this line with the Expect line itself? interrupts readability.

rogpeppe: Because this is the way it's done in every other test and this PR is about changing exactly one property of a couple of tests, not changing things which are technically fine the way they are now and might have been done for good reason.

If we were to change other aspects of these tests, there are way more important things to do, in my view (see #416 for an example).

jonjohnsonjr (approved):

PR #430, opened 2023-06-26
conformance: make references tests part of content discovery workflow

jdolitsky:


sudo-bmitch (approved):

jonjohnsonjr (approved):

conformance/03_discovery_test.go
```diff
@@ -82,9 +84,151 @@ var test03ContentDiscovery = func() {
...
+				Expect(resp.StatusCode()).To(SatisfyAll(
+					BeNumerically(">=", 200),
+					BeNumerically("<", 300)))
+				Expect(resp.Header().Get("OCI-Subject")).To(Equal(manifests[4].Digest))
+
+				// Populate registry with test references manifest (ArtifactType, config.MediaType = emptyJSON)
+				req = client.NewRequest(reggie.PUT, "/v2/<name>/manifests/<reference>",
+					reggie.WithReference(refsManifestALayerArtifactDigest)).
+					SetHeader("Content-Type", "application/vnd.oci.image.manifest.v1+json").
+					SetBody(refsManifestALayerArtifactContent)
```
andaaron, line 140: Hi @jdolitsky

Looking at this test @peusebiu and I noticed `refsManifestALayerArtifactContent` actually contains a reference to layer `testRefBlobB` (see https://github.com/opencontainers/distribution-spec/blob/main/conformance/setup.go#L396), which is not yet uploaded to the registry.

My understanding of the push specifications is that layer blobs should be pushed by the client before the manifest.
If the server decides to reject them, it remains fully compliant. So why would the test check the server accepts this upload.

We're not sure if this is by design of the test, or a bug, as `testRefBlobB` is uploaded later at line 198.
On the other hand the code in setup.go is strange, maybe `testRefBlobA` should have been reference there is setup.go at line 396

jdolitsky: Yes, I defer to Ram. In this PR I tried to just copy-paste his work

PR #449, opened 2023-07-21
conformance: OCI-Filters-Applied should return a literal

rchincha:
Per spec, the filter type (in this case "artifactType") should be returned instead of the actual filter values.

Fixes https://github.com/opencontainers/distribution-spec/issues/448

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ conformance-test:
 
 conformance-binary: $(OUTPUT_DIRNAME)/conformance.test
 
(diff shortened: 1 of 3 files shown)


jdolitsky (dismissed):

jdolitsky:
in the same boat as https://github.com/opencontainers/distribution-spec/pull/436 with CI failing - do we have a zot release/rc yet that will go green for these changes?

sudo-bmitch (approved):
LGTM

sudo-bmitch (approved):

jdolitsky (approved):

PR #453, opened 2023-07-27
Rewrite conformance action to enable locking to spec version

jdolitsky:


jdolitsky:
cc @rchincha - I know y'all are using this action for zot. Let me know if any concerns. This should allow you to do things like the following:

```
- name: Test OCI conformance v1.0
  uses: opencontainers/distribution-spec@v1.0.1

- name: Test OCI conformance v1.1
  uses: opencontainers/distribution-spec@v1.1.0
```

dmcgowan (approved):

mikebrow (approved):
LGTM

PR #486, opened 2023-10-30
Introduce OCI-Chunk-Max-Length header in the docs

gabivlj:
Related to: https://github.com/opencontainers/distribution-spec/issues/485

This allows registries that buffer request bodies via proxies to be able to handle uploads that surpass their in-memory limit.


sudo-bmitch (changes requested):
Let's hold on this PR pending the discussion on #485.

PR #537, opened 2024-05-29
Add support for content range requests when getting blobs

### LINKED ISSUE #354: How resumable pulls should work?
The spec talks about [Resumable Pull](https://github.com/opencontainers/distribution-spec/blob/v1.0.1/spec.md#resumable-pull) but does not specify how this should be done. The Docker spec states the following in the section [Pulling a Layer](https://github.com/opencontainers/distribution-spec/blob/v1.0.1/spec.md#resumable-pull)

To allow for incremental downloads, `Range` requests should be supported, as well.

This is also later in the spec defined in more detail in the section [Fetch Blob Part](https://github.com/distribution/distribution/blob/5cb406d511b7b9163bff9b6439072e4892e5ae3b/docs/spec/api.md#fetch-blob-part).

rchincha:
OCI artifacts support has landed in various OCI specs v1.1.0 which allows for arbitrary artifact types, small and large.

Large artifacts (even existing container images) pose a particular challenge that:
1.    it takes too long to download
2.   it takes too long to unpack

This PR begins to address 1. above.

The client can initiate a HEAD request to get the size and later multiple GET range requests to download a blob in parallel.

Fixes #354

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -190,6 +190,8 @@ If present, the value of this header MUST be a digest matching that of the respo
 
 If the blob is not found in the registry, the response code MUST be `404 Not Found`.
 
+A registry SHOULD support the `Range` request header in accordance with [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-range-requests).
+
 ##### Checking if content exists in the registry
 
 In order to verify that a repository contains a given manifest or blob, make a `HEAD` request to a URL in the following form:


sudo-bmitch:
Does this need to be included in the dist-spec, vs delegating to the HTTP 1.1 RFC: https://datatracker.ietf.org/doc/html/rfc7233#section-3.1

rchincha:
https://github.com/opencontainers/distribution-spec/issues/536

rchincha:
> Does this need to be included in the dist-spec, vs delegating to the HTTP 1.1 RFC: https://datatracker.ietf.org/doc/html/rfc7233#section-3.1

If it is not included here, registry operators have no guidance on whether to implement this or not. We are explicitly stating that we are doing/requiring this for perf reasons, etc.

spec.md
```diff
@@ -190,6 +190,11 @@ If present, the value of this header MUST be a digest matching that of the respo
 
 If the blob is not found in the registry, the response code MUST be `404 Not Found`.
 
+A GET request may also include a `Range` request header to download part of a
+blob. The response code can either be `216 (Partial Content)` or `416 (Range
+Not Satisfiable)` in case of an invalid range. If the registry doesn't support
+range requests, it can respond with `Accept-Ranges: none`.
```
samuelkarp, line 196:
```suggestion
Not Satisfiable)` in case of an invalid range. A registry MAY ignore the `Range` header.
```

samuelkarp (commented):
If we recommend behavior here, that should align with both the Notational Conventions of this specification and with the behavior described in RFC 9110.

spec.md
rchincha: An older registry unaware of this change may ignore the header, but a newer registry may want to send a response back to discourage clients from send the Range request.

rchincha:
We can only download as fast as the pipe allows of course, but maybe this helps with head-of-line blocking when transferring a large file by getting many smaller pieces in parallel - how many pieces? etc is tbd math.

rchincha:
Updated as per comments in the OCI meeting 05/30/2024
The only point of contention then is MAY vs SHOULD 

Candidate for dist-spec v1.1.1?

spec.md
```diff
@@ -190,6 +190,9 @@ If present, the value of this header MUST be a digest matching that of the respo
 
 If the blob is not found in the registry, the response code MUST be `404 Not Found`.
 
+A GET request MAY also include a `Range` request header to download part of a
+blob in accordance with [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-range-requests).
```
sudo-bmitch, line 194: I reworded it to focus on the registry since "a request MAY include" could imply "a registry MUST support" which I don't think is the goal.

Also a couple nits:

- One sentence per line for easier PRs (remove the linefeed here).
- I think we should standardize on a single RFC host. It looks like most of the links here use https://www.rfc-editor.org/.

```suggestion
A registry SHOULD support the `Range` request header in accordance with [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-range-requests).
```

sudo-bmitch (changes requested):
In addition to the reword, it looks like a separate commit got pulled in by accident. Please fix with a rebase on main.

sudo-bmitch (approved):
LGTM

rchincha:
Note that another reading of the spec found this:
https://github.com/opencontainers/distribution-spec/blob/main/spec.md#resumable-pull
So we are just making this explicit with this PR.

sudo-bmitch:
I've linked with #354.

rchincha:
Just preliminary testing ... 

zot localhost over http with 10GiB artifact blob

Single-threaded client start to finish: 1min 32secs 

{"level":"info","module":"http","component":"session","clientIP":"127.0.0.1:59934","method":"GET","path":"/v2/artifact/blobs/sha256:732377e7f4a2abdc13ddfa1eb4c9c497fd2a2b294674d056cf51581b47dd586d","statusCode":200,**"latency":"1m32s"**,"bodySize":10737418240,"headers":{"Accept-Encoding":["gzip"],"User-Agent":["Go-http-client/1.1"]},"goroutine":167,"caller":"zotregistry.dev/zot/pkg/api/session.go:132","time":"2024-06-08T03:34:08.916403767Z","message":"HTTP API"}

Multi-theaded client (still HTTP/1.1) but with range based download: ~33 secs
HTTP/1.1 doesn't do true pipelining, however at least, the server backend becomes aware of all requests so can start working on I/Os in parallel. golang HTTP/2.0 needs TLS as a default (without doing transport gymnastics) - will test this next.

{"level":"info","module":"http","component":"session","clientIP":"127.0.0.1:52134","method":"GET","path":"/v2/artifact/blobs/sha256:732377e7f4a2abdc13ddfa1eb4c9c497fd2a2b294674d056cf51581b47dd586d","statusCode":206,**"latency":"18s"**,"bodySize":2147483649,**"headers":{"Range":["bytes=6442450944-8589934592"]**,"User-Agent":["Go-http-client/1.1"]},"goroutine":184,"caller":"zotregistry.dev/zot/pkg/api/session.go:132","time":"2024-06-08T04:11:01.590093476Z","message":"HTTP API"}
{"level":"info","module":"http","component":"session","clientIP":"127.0.0.1:52176","method":"GET","path":"/v2/artifact/blobs/sha256:732377e7f4a2abdc13ddfa1eb4c9c497fd2a2b294674d056cf51581b47dd586d","statusCode":206,"latency":"18s","bodySize":2147483648,"headers":{"Range":["bytes=8589934592-10737418240"],"User-Agent":["Go-http-client/1.1"]},"goroutine":195,"caller":"zotregistry.dev/zot/pkg/api/session.go:132","time":"2024-06-08T04:11:01.831432295Z","message":"HTTP API"}

rchincha:
image_720

@shizhMSFT reported this for Azure. Thanks!

shizhMSFT:
> image_720
> 
> @shizhMSFT reported this for Azure. Thanks!

Here's the detailed data points for the above diagram.

| Connections | Download Speed (MiB/s) |
| ----------- | ---------------------- |
| 1           | 132                    |
| 2           | 142                    |
| 3           | 147                    |
| 4           | 144                    |
| 5           | 145                    |
| 6           | 143                    |
| 7           | 147                    |
| 8           | 142                    |
| 9           | 145                    |
| 10          | 142                    |
| 11          | 146                    |
| 12          | 141                    |
| 13          | 143                    |
| 14          | 140                    |
| 15          | 144                    |
| 16          | 144                    |

The benchmark was generated by running `for i in {1..16}; do aria2c -x 16 -s $i $url; done` in an Azure `D2s v3` VM with Premium SSD to download a 10 GiB blob in an Azure Blob Storage instance within the same region.

Detailed conversation can be found in the CNCF Slack channel #oras: https://cloud-native.slack.com/archives/CJ1KHJM5Z/p1718311259363889

dmcgowan (approved):

PR #568, opened 2025-02-20
Add optional tagType query param to tags list endpoint

jdolitsky:
Adds the ability to do a server-side filter of "digest" tags (e.g. `sha256-...`).

This is done by creating a `tagType` classification, which can be used to filter things. This works similar to the `artifactType` filter on the referrers API.

This is intended to exclude or isolate tags which are considered "non-normal". If another type of common non-normal tag appears in the wild, it can be added as a new `tagType`. In general, this is considered an anti-pattern and clients are encouraged to use the `subject` field and referrers API etc. to connect objects in the registry.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -559,6 +559,17 @@ The tags MUST be in lexical order.
 
 When using the `last` query parameter, the `n` parameter is OPTIONAL.
 
+The registry SHOULD support filtering on `tagType`.
+To fetch the list of tags with a filter, perform a `GET` request to a path in the following format: `/v2/<name>/tags/list?tagType=<tagType>` <sup>[end-8c](#endpoints)</sup>
+If filtering is requested and applied, the response MUST include a header `OCI-Filters-Applied: tagType` denoting that an `tagType` filter was applied.
+If multiple filters are applied, the header MUST contain a comma separated list of applied filters.
+If the `OCI-Filters-Applied` is included, the `n` and `last` query parameters and any `Link` headers MUST be applied
+to the filtered list of tags.
+
+If the `tagType` query parameter is provided, it MUST be one of the following values:
+- `digest`: tags values which start with the prefix `sha256-` used in [Referrers Tag Schema](#referrers-tag-schema) and elsewhere
+- `real`: tag values which do not match another `tagType` listed above (e.g. "latest")
+
 _Implementers note:_
 Previous versions of this specification did not include the `Link` header.
 Clients depending on the number of tags returned matching `n` may prematurely stop pagination on registries using the `Link` header.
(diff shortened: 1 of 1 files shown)


tianon:
"this is considered an anti-pattern" seems like a pretty good reason _not_ to add even more functionality supporting this in the spec, right? Especially given the long tail of adoption?

jdolitsky:
> "this is considered an anti-pattern" seems like a pretty good reason not to add even more functionality supporting this in the spec, right? Especially given the long tail of adoption?

That's a fair point. However, specifically in the case of tags falling into this "digest" category, we're only seeing the number of these tags grow vs. slow down/migrating to use referrers.

jdolitsky:
From discussion, alternative:

- Request tag endpoint with `Accept: <new_media_type>`
- If `Content-Type` of response has this mediatype, parse it new way
- Query params for search query regex and artifact type etc.
- Support for ETAG header

jcarter3:
If implementers are resource constrained on adding referrers, it seems very unlikely for something like this to get implemented either.

FWIW, Docker Hub does have referrers on the roadmap, and would not likely prioritize this ahead of that.

jcarter3:
What about cosign style tags with a `.sig` or `.att` suffix? Are those considered "digest" style? They aren't valid digests, nor referrers, but are "not real".

jdolitsky:
> If implementers are resource constrained on adding referrers, it seems very unlikely for something like this to get implemented either.

Fair point. I do think that depends. There was further discussion on the call today that there are other use cases for filtering tags unrelated to referrers-related behaviour:

- Filtering out non-semver
- Filtering out semver `rc` tags
- "Give me only things pushed in last day" (not currently possible in current spec)

> What about cosign style tags with a .sig or .att suffix? Are those considered "digest" style? They aren't valid digests, nor referrers, but are "not real".

Yes, the `.att` / .`sig` tags are intended to fall into "digest" the category. As currently written in this PR, it applies to any tag prefixed with the string "sha256-" (we could also have it apply to "sha512-" and "blake-" if necessary). A regex filter is another possibility vs. known tag types.

jcarter3:
I agree that other types of tag listing would be useful, but that seems much greater in scope. Current spec limits the response to be a string list of tags, which severely limits the usefulness. If we want to do date filters, it'd be more useful to be able to return some metadata _with_ the tag, which would be a new endpoint (or add new, optional fields in the response to keep backwards compatibility)

jdolitsky:
@jcarter3 - see this comment above for @sudo-bmitch rough idea for a v2 tag listing endpoint: https://github.com/opencontainers/distribution-spec/pull/568#issuecomment-2672337832

Issue #506, opened 2024-01-30
Question regarding conformance tests for push


ViddeM:
Hello! 

I'm not sure if I'm just confused but it seems to me that the conformance test for push doesn't quite match the specification.
The only PATCH endpoint in the spec (as far as I understand) is for the chunked upload flow ([end-5](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints)) which seems to require the three headers `Content-Type`, `Content-Range` and `Content-Length`. However, in the conformance test for `Blob upload streamed` ([link](https://github.com/opencontainers/distribution-spec/blob/05cf62402d050b09953f4ed9a041666d150d8718/conformance/02_push_test.go#L33)) there is a patch request that only provides the `Content-Type` header. 

Am I confused or is this flow not covered by the specification and if so, why is it in the conformance tests?

ViddeM:
Yeah I sorta started to assume that after doing some more digging, I found that Microsofts container registry API documentation seems to cover that flow in (enough) detail to implement something that passes the conformance tests.

https://learn.microsoft.com/en-us/rest/api/containerregistry/blob/upload

PR #576, opened 2025-05-30
fix: add required Content-Range header to PATCH test

coryodaniel:
The spec requires Content-Range header for PATCH requests, but the conformance test "PATCH request with blob in body should yield 202 response" omits this requirement. This creates a mismatch between the spec and the test suite.

The spec clearly states that for PATCH requests:
- Content-Range is required
- Must be inclusive on both ends
- First chunk must begin with 0
- Must match regex ^[0-9]+-[0-9]+$

This fix adds the required Content-Range header to the test, ensuring it properly validates spec compliance.

I implemented this a little hacky to illustrate, I wanted to make sure I wasn't misinterpreting the spec, but when I enforced the content-range header on PATCH in our [registry](https://github.com/massdriver-cloud/oci), it caused the conformance test suite to fail.

Happy to move stuff into setup.go if you want it to be inline with the other setup variables. I didn't do it there since its not a "chunk" and that appears to be the naming convention. Anywho, thanks!

### DIFF
--- a/conformance/02_push_test.go
+++ b/conformance/02_push_test.go
@@ -30,8 +30,12 @@ var test02Push = func() {
 				location := resp.Header().Get("Location")
 				Expect(location).ToNot(BeEmpty())
 
+				length, _ := strconv.Atoi(testBlobALength)
+
 				req = client.NewRequest(reggie.PATCH, resp.GetRelativeLocation()).
 					SetHeader("Content-Type", "application/octet-stream").
+					SetHeader("Content-Length", testBlobALength).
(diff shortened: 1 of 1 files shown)


sudo-bmitch (changes requested):
This is the streaming upload, which we do not want to change to a chunked upload (that's already a separate test). See #506 for details. It's a deficiency in the spec, not in the conformance test.

dmcgowan:
Clarification in the spec is reasonable, removing from the conformance test is not the ideal solution though.

PR #579, opened 2025-06-26
Initial PR to propose updates to tag listing

jcarter3:
Motivation:
Provide registry agnostic mechanism to get new tags

Related to discussions at:
* https://github.com/opencontainers/distribution-spec/issues/575
* https://github.com/opencontainers/distribution-spec/pull/568

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -559,8 +559,9 @@ The tags MUST be in lexical order.
 
 When using the `last` query parameter, the `n` parameter is OPTIONAL.
 
-Registries MAY support an enhanced tag listing experience when clients pass an `Accept` header value of 
-`application/vnd.oci.image.index.v1+json`. If a registry supports this mode, they MUST support these additional query parameters:
+Registries MAY support an enhanced tag listing experience when clients pass an `Accept` header value of `application/vnd.oci.image.index.v1+json`.
+If a registry supports this mode, they MUST set the `Content-Type: application/vnd.oci.image.index.v1+json` header.
+Registries supporting the enhanced tag list MUST support these additional query parameters:
 * `sort_by` - allowed values are `created_at`, `name`
 * `sort_order` - allowed values are `asc` and `desc`
 
@@ -593,7 +594,7 @@ The response MUST be a json body in the following format:
 ```
 The following annotations MUST be included:
 * `org.opencontainers.image.ref.created` - Date and time the reference was created (date-time string as defined by RFC 3339). If unknown, default to epoch.
-* `org.opencontainers.image.ref.name` - Reference name of the manifest.
+* `org.opencontainers.image.ref.name` - Name of the tag.
 
 The following annotations MAY be included:
 * `org.opencontainers.image.created` - Date and time on which the image was built (date-time string as defined by RFC 3339).
@@ -603,8 +604,8 @@ The following annotations MAY be included:
 _Implementers note:_
 Previous versions of this specification did not include the `Link` header.
 Clients depending on the number of tags returned matching `n` may prematurely stop pagination on registries using the `Link` header.
-When available, clients should prefer the `Link` header over using the `last` parameter for pagination. If using enhanced
-tag listing, the `Link` header MUST be provided.
+When available, clients should prefer the `Link` header over using the `last` parameter for pagination.
+If using enhanced tag listing, clients MUST use the `Link` header for pagination.
 
 ##### Listing Referrers
 
--- a/spec.md
+++ b/spec.md
@@ -559,10 +559,52 @@ The tags MUST be in lexical order.
 
 When using the `last` query parameter, the `n` parameter is OPTIONAL.
 
+Registries MAY support an enhanced tag listing experience when clients pass an `Accept` header value of 
+`application/vnd.oci.image.index.v1+json`. If a registry supports this mode, they MUST support these additional query parameters:
+* `sort_by` - allowed values are `created_at`, `name`
+* `sort_order` - allowed values are `asc` and `desc`
+
+The response MUST be a json body in the following format:
+```json
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.index.v1+json",
+  "manifests": [
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -559,10 +559,52 @@ The tags MUST be in lexical order.
...
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.index.v1+json",
+  "manifests": [
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "size": 1234,
+      "digest": "sha256:a1a1a1...",
+      "annotations": {
+        "org.opencontainers.image.ref.created": "2022-01-01T14:42:55Z",
+        "org.opencontainers.image.ref.name": "<tag1>"
```
qweeah, line 579: If a manifest contains several tags, would it be more efficient to merge all the tags into one entry here?

Wwwsylvia, line 579: Do these annotations originate from the original manifest data, or are they generated by the registry and added to the descriptors? 🤔

- If they come from the manifest data, how does the API handle manifests that do not contain these annotations? 
- If they are generated by the registry, how does the API manage potential conflicts between the manifest annotation values and the registry-recorded values, such as creation time?

jcarter3: They are generated by the registry. The registry may optionally include other annotations, including those from the manifest if it chooses.

For conflicts, there _shoudn't_ be any. The `...image.ref.*` tags are specific to a reference, not to a manifest - As in, a manifest shouldn't include an annotation of the reference it will be tagged with _after_ it's pushed. Though maybe there is some weird edge case where this happens :thinking:

sudo-bmitch (changes requested):
We need to add conformance tests, and get some example implementations.

Edit to add: if we add more sort orders or filters in the future, how would the client know if they were applied or if the registry only supports the first iteration of these features?

spec.md
```diff
@@ -559,10 +559,53 @@ The tags MUST be in lexical order.
...
+      "annotations": {
+        "org.opencontainers.image.ref.created": "2022-01-01T07:21:33Z",
+        "org.opencontainers.image.ref.name": "<tag2>"
+      }
+    }
+  ]
+}
+```
+The following annotations MUST be included:
+* `org.opencontainers.image.ref.created` - Date and time the reference was created (date-time string as defined by RFC 3339). If unknown, default to epoch.
```
shizhMSFT, line 596: A resource usually has two timestamps: one for creation and another for modification. For example, tag `v1` is created on `2022-01-01T07:21:33Z` and later re-tagged on `2022-02-03T08:21:21Z`. Do you want to introduce a new annotation for the modification / update time?

spec.md
```diff
@@ -559,10 +559,53 @@ The tags MUST be in lexical order.
 
 When using the `last` query parameter, the `n` parameter is OPTIONAL.
 
+Registries MAY support an enhanced tag listing experience when clients pass an `Accept` header value of `application/vnd.oci.image.index.v1+json`.
+If a registry supports this mode, they MUST set the `Content-Type: application/vnd.oci.image.index.v1+json` header.
+Registries supporting the enhanced tag list MUST support these additional query parameters:
+* `sort_by` - allowed values are `created_at`, `name`
+* `sort_order` - allowed values are `asc` and `desc`
+
+The response MUST be a json body in the following format:
```
shizhMSFT, line 568: Just a note: This example response looks like an `index.json` in the OCI image layout, which serves as a tag list or generally a manifest list.

spec.md
jcarter3: I'd say this could be an optional value added by registries, but we can specify an annotation for it. Some registries may not have this information, and it might be of little value (it wouldn't give a history) on it's own.

spec.md
tianon: That's entirely intentional, and even why it's reusing the annotation that's "only for `index.json`" 😅

Reusing the "image index" structure for this also matches the design of the referrers API, which is a useful pattern because it avoids defining "yet another data schema" without a really strong reason.

spec.md
```diff
@@ -559,10 +559,52 @@ The tags MUST be in lexical order.
 
 When using the `last` query parameter, the `n` parameter is OPTIONAL.
 
+Registries MAY support an enhanced tag listing experience when clients pass an `Accept` header value of 
+`application/vnd.oci.image.index.v1+json`. If a registry supports this mode, they MUST support these additional query parameters:
+* `sort_by` - allowed values are `created_at`, `name`
+* `sort_order` - allowed values are `asc` and `desc`
```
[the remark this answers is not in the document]
Ducbiane: Vậy phải làm sau ạ

Ducbiane:
Đã thay đổi sdt củ bằng số đt trên số fone đã đăng ký và cũng thay mật khẩu trên id đó thay luôn gmail nộp hết giấy tờ liên quang gần 1 tháng nay ngày nào cũng làm xác minh nhưng ko đc hỗ trợ và bị từ chối

Ducbiane:
[Uploading Screenshot_2025-09-03-06-19-57-44_ee1cec40dcf6eb3919ecbfc5d87e6719.jpg…]()
[Screenshot_2025-09-03-06-21-23-58_ee1cec40dcf6eb3919ecbfc5d87e6719](https://github.com/user-attachments/assets/1be29bd1-dbc8-4429-b8af-695cdf481a74)
Screenshot_2025-09-03-06-21-15-75_ee1cec40dcf6eb3919ecbfc5d87e6719
Screenshot_2025-09-03-06-20-56-72_ee1cec40dcf6eb3919ecbfc5d87e6719
Screenshot_2025-09-03-06-21-06-61_ee1cec40dcf6eb3919ecbfc5d87e6719
+ Index "Json"

spec.md
tianon: While discussing #606 in today's call, we talked about this same problem, and we noted that over there, we could trim anything that starts with `org.opencontainers.distrubution.` to solve this problem (so that future additions are clearly owned by the distribution spec).

The exact same solution doesn't work here as-is, but I personally think `org.opencontainers.image.ref.` would be a very reasonable namespace for this to trim (and it could trim `org.opencontainers.distrubution.` too, for good measure, if we were worried about future expansion).

PR #595, opened 2026-01-21
Make Docker-Content-Digest non-optional

syed:
Based on the discussion on the call, we want to have `Docker-Content-Digest` as a required header when doing a HEAD on a reference. This fixes an issue where given a tag for a manifest, there is no way currently for the client to know which hashing algorithm was used when pushing

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -48,8 +48,8 @@ For relevant details and a history leading up to this specification, please see 
 
 #### Legacy Docker support HTTP headers
 
-Because of the origins of this specification, the client MAY encounter Docker-specific headers, such as `Docker-Content-Digest`, or `Docker-Distribution-API-Version`.
-These headers are OPTIONAL and clients SHOULD NOT depend on them.
+Because of the origins of this specification, the client MAY encounter Docker-specific headers, such as `Docker-Distribution-API-Version`.
+Unless documented elsewhere in the spec, these headers are OPTIONAL and clients SHOULD NOT depend on them.
 
 #### Legacy Docker support error codes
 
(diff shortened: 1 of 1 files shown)


tianon (approved):
LGTM (non-binding, obviously)

sudo-bmitch (commented):
Given the goal from our conversation is to make it possible to verify the digest of a tag to attach referrers and pin references, I'm also looking at other parts of the spec. Specifically, in a few places we have:

> The `Docker-Content-Digest` header, if present on the response, returns the canonical digest ...

Our usage of "canonical digest" references sha256, so I think the language there should be adjusted to reference the pushed digest when available. Though perhaps we want to save that change for once we have a way to push a tagged manifest with a different digest algorithm.

mikebrow (commented):
for the purposes of documenting the issue outside the call.. 

The originally added text "These headers are OPTIONAL and clients SHOULD NOT depend on them." Appears to have been written in error and only meant to cover the version header and was not meant for the digest.

see suggested changes.

sudo-bmitch (changes requested):
Overall, the SHOULD -> MUST changes LGTM. I don't think we ever define "canonical digest" anywhere, perhaps we should have an entry in the definitions.

dmcgowan (dismissed):

tianon (approved):

sudo-bmitch (changes requested):
One line needs to get added back. And we need to squash the commits. After that, it LGTM.

sudo-bmitch:
Thanks. Please [squash the commits](https://stackoverflow.com/q/5189560/596285) and I'll give it my 👍🏼 .

syed:
squashed

sudo-bmitch:
> squashed

Looks like the DCO signoff got lost (or maybe wasn't there). `git rebase HEAD~1 --signoff`

sudo-bmitch (approved):
LGTM!

dmcgowan (approved):

andaaron:
Is the `Docker-Content-Digest` a MUST also on status code 206 (for https://github.com/opencontainers/distribution-spec/blob/main/spec.md?plain=1#L196 support, that could result in a contradiction of https://github.com/opencontainers/distribution-spec/blob/main/spec.md?plain=1#L190 which mentions the value should match the digest of the body in the current response)?

Offtopic: the spec doesn't specify code 206 at https://github.com/opencontainers/distribution-spec/blob/main/spec.md?plain=1#L809, but https://www.rfc-editor.org/rfc/rfc9110.html#name-206-partial-content covers it.

sudo-bmitch:
@andaaron my personal opinion is that clients should be verifying the digest that they requested on a blob pull, which is why we say https://github.com/opencontainers/distribution-spec/blob/main/spec.md?plain=1#L192:

> Clients SHOULD verify that the response body matches the requested digest.

So I ignore the registry header, and depend on what I requested instead. This prevents some security vulnerabilities where a client requests one digest and the registry replies with a different one.

The big reason for this PR was that clients need to know the digest algorithm the registry associates with a manifest when pinning content, creating an artifact with a subject, and fetching artifacts with the referrers API. Without that, the workflow for changing the digest algorithm was broken. It also cleaned up other parts of the code for consistency which is what I think you're looking at.

If you're interested in cleaning up some edge cases around partial pulls, feel free to send over a PR for that. I think we all agree the header needs to match the digest of the entire blob, and not of the single partial body response. We had a similar issue on the chunked push that was clarified on https://github.com/opencontainers/distribution-spec/blob/main/spec.md?plain=1#L400:

> The closing `PUT` request MUST include the `<digest>` of the whole blob (not the final chunk) as a query parameter.

PR #597, opened 2026-01-28
Improve manifests cleanup

Silvanoc:
Running the tests against a local ZOT registry it is possible to observe that the tests don't delete properly all pushed manifests and blobs. This PR introduces a list to keep track of pushed manifests and blobs to properly clean them up on test termination.

Additionally reducing code duplication through the introduction of functions for pushing and deleting manifests and blobs.


sudo-bmitch:
Closing since the conformance tests have been redesigned.

PR #601, opened 2026-03-26
Remove the ignore on go.sum

sudo-bmitch:
Ignoring the go.mod and go.sum required reverting the ignore in sub-directories. Instead, this is a good use case for a temporary go.work. This also updated the pandoc container to use the upstream image, aligning with image-spec.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 output
 header.html
 tags
-go.mod
-go.sum
+go.work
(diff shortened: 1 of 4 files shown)


jdolitsky (approved):

tianon (approved):
(non-binding, non-maintainer)

dmcgowan (approved):

PR #606, opened 2026-04-23
Update spec to include tag history endpoint

jcarter3:
Updates distribution spec to include new endpoint for retrieving the history of a given tag. The response is a json list of descriptors,including the created timestamp and sorted in descending order (newest first).
The list can be paged.


spec.md
```diff
@@ -601,6 +601,39 @@ Previous versions of this specification did not include the `Link` header.
 Clients depending on the number of tags returned matching `n` may prematurely stop pagination on registries using the `Link` header.
 When available, clients should prefer the `Link` header over using the `last` parameter for pagination.
 
+#### Listing Tag History
+
+To fetch the history of a given tag, perform a `GET` request to a path in the following format: `/v2/<name>/tag/<tag>/history` <sup>[end-15](#endpoints)</sup>
```
sudo-bmitch: If someone called:

```
/v2/project-x/tag/manifests/history
```

That could be the tag history for the tag `manifests`. Or it could be the manifest for the tag `history` in the repo `project-x/tag`.

sudo-bmitch: One idea is to use the `_oci` extension: `_oci/tags/history`.

extensions/_oci.md
```diff
@@ -66,6 +66,106 @@ Content-Type: application/json
...
+
+##### Query Parameters
+
+The following query parameters MAY be provided:
+
+- **`n`** *integer*, OPTIONAL
+
+  Specifies the maximum number of results to return.
+  The registry MAY return fewer results than requested if fewer historical entries exist.
+  When `n` is zero, this endpoint MUST return an empty array.
```
mikebrow:
```suggestion
  When `n` is zero, this endpoint MUST return an empty array, useful for determining if history is available (200) or not (404) for the tag query.
```

extensions/_oci.md
```diff
@@ -66,6 +66,106 @@ Content-Type: application/json
...
+  When `n` is not specified, the registry MAY apply a default limit.
+
+- **`before`** *string (RFC 3339 timestamp)*, OPTIONAL
```
mikebrow, line 179: maybe from or ot (older than)
 

```suggestion
- **`ot`** *string (RFC 3339 timestamp)*, OPTIONAL
```

extensions/_oci.md
```diff
@@ -66,6 +66,106 @@ Content-Type: application/json
...
+
+  When provided, only entries whose `org.opencontainers.tag.created` value is strictly less than (i.e. older than) `before` will be returned.
```
mikebrow:
```suggestion
  When provided, only entries whose `org.opencontainers.tag.created` value is strictly less than (i.e. older than) the `ot` timestamp will be returned.
```

extensions/_oci.md
mikebrow, line 179: is there a chance to have two results with the exact same time stamp..

extensions/_oci.md
```diff
@@ -66,6 +66,106 @@ Content-Type: application/json
...
+  This is used for time-based pagination: pass the `org.opencontainers.tag.created` value of the last entry returned in the previous response to retrieve the next page.
+
```
mikebrow, line 193: another idea would be to include a "newer-than" option, or a "between" timestamp1/timestamp2 allowing a client to query new tags or a time range..

mikebrow (commented):
some thoughts

extensions/_oci.md
```diff
@@ -66,6 +66,148 @@ Content-Type: application/json
...
+### Component: `tag-history`
+
+This component is for endpoints relating to tag history operations on a repository.
+
+This endpoint returns the history of a tag, listing each manifest the tag has pointed to over time and each time the tag was deleted, in descending order by history entry timestamp (newest first).
+
+Tag history MAY be retrieved with a standard `GET` as follows.
+
+```HTTP
+GET /v2/<name>/_oci/tag-history/<tag>
```
sudo-bmitch, line 78: I think `<component>=tag`, `<module>=history`, and use parameters for passing the tag, better follows the extensions syntax documented at https://github.com/opencontainers/distribution-spec/tree/main/extensions#repository-level-extensions

```suggestion
GET /v2/<name>/_oci/tag/history/?tag=<tag>
```

(Note that I personally wish we weren't so prescriptive and only specified the extension is `_<extension>` with everything after that being implementation defined, but I suspect I lost that battle when we added extensions to the spec.)

extensions/_oci.md
```diff
@@ -66,6 +66,148 @@ Content-Type: application/json
...
+```
+
+`<name>` is the namespace of the repository, and `<tag>` is the name of the tag whose history is being queried.
+
+A successful request MUST return a `200 OK` response code.
+If the repository does not exist, the registry MUST return a `404 Not Found` response code.
+If the tag has no history, the registry MUST return a `404 Not Found` response code.
```
sudo-bmitch:
```suggestion
If the tag has no history or the registry does not implement this extension, the registry MUST return a `404 Not Found` response code.
```

extensions/_oci.md
```diff
@@ -66,6 +66,148 @@ Content-Type: application/json
...
+Tag history SHOULD remain queryable after the tag is deleted.
+Deleted manifests MUST NOT be removed from tag history.
```
sudo-bmitch: MUST feels a bit strong here, and may break some retention and redaction requirements of some registries.

```suggestion
Deleted manifests SHOULD NOT be removed from tag history.
```

extensions/_oci.md
```diff
@@ -66,6 +66,148 @@ Content-Type: application/json
...
+History entries for deleted manifests MUST continue to include the descriptor fields recorded when the tag was assigned to the manifest.
```
sudo-bmitch: Is this rephrasing the previous line? If so, I'd consolidate the two, with a "SHOULD".

extensions/_oci.md
```diff
@@ -66,6 +66,148 @@ Content-Type: application/json
...
+        "size": 1234,
+        "digest": "sha256:b2b2b2...",
+        "annotations": {
+            "org.opencontainers.tag.created": "2026-04-03T02:01:00Z"
+        }
+    }
+]
+```
+
+Results MUST be sorted in descending order by the history entry timestamp, which is either the `org.opencontainers.tag.created` annotation value or the `org.opencontainers.tag.deleted` annotation value (i.e. the most recent entry appears first).
```
sudo-bmitch: In image spec, we use the prefix `org.opencontainers.image`. For distribution, perhaps we should standardize on `org.opencontainers.distribution` to avoid conflicts with other specs?

extensions/_oci.md
```diff
@@ -66,6 +66,148 @@ Content-Type: application/json
...
+
+Each descriptor object in the response MUST be either a manifest descriptor or a delete marker.
```
sudo-bmitch:
```suggestion
Each descriptor object in the response MUST be either a create or delete entry.
```

extensions/_oci.md
```diff
@@ -66,6 +66,148 @@ Content-Type: application/json
...
+
+Manifest descriptors MUST include the following properties:
```
sudo-bmitch, line 139:
```suggestion
Tag create entries, which includes replacing an existing tag, MUST include the following properties:
```

extensions/_oci.md
```diff
@@ -66,6 +66,148 @@ Content-Type: application/json
...
+
+- **`annotations`** *map of strings*, REQUIRED
+
+  Annotations associated with the historical tag entry. MUST include:
+
+  - **`org.opencontainers.tag.created`** *string*, REQUIRED
+
+    The RFC 3339 timestamp at which the tag was assigned to this manifest. Used as the sort key and as the cursor for time-based pagination.
+
+Delete markers MUST be represented by the OCI empty descriptor and MUST include the following properties:
```
sudo-bmitch:
```suggestion
Tag delete entries MUST include the following properties:
```

extensions/_oci.md
```diff
@@ -66,6 +66,148 @@ Content-Type: application/json
...
+
+  MUST be `e30=`.
+
+- **`annotations`** *map of strings*, REQUIRED
+
+  Annotations associated with the historical tag entry. MUST include:
+
+  - **`org.opencontainers.tag.deleted`** *string*, REQUIRED
+
+    The RFC 3339 timestamp at which the tag was deleted. Used as the sort key and as the cursor for time-based pagination.
```
sudo-bmitch: I'd rephrase to use the descriptor of the deleted entry. And now with so much overlap, I think we can probably merge the two create/delete sections and then document the two possible annotations.

```suggestion
- **`mediaType`** *string*, REQUIRED

  The media type of the manifest this tag previously pointed to at that point in time.

- **`size`** *integer*, REQUIRED

  The size in bytes of the manifest.

- **`digest`** *string*, REQUIRED

  The digest of the manifest, in the form `<algorithm>:<encoded>`.

- **`annotations`** *map of strings*, REQUIRED

  Annotations associated with the historical tag entry. MUST include:

  - **`org.opencontainers.distribution.tag.deleted`** *string*, REQUIRED

    The RFC 3339 timestamp at which the tag was deleted. Used as the sort key and as the cursor for time-based pagination.
```

sudo-bmitch (commented):
Here's some early feedback. I still need to spend some time implementing this before considering approving, and we need a conformance test.

extensions/_oci.md
jcarter3: How hard is it to push back and change this? That design makes every downstream extension significantly worse. If there aren't any defined extensions yet, it should be possible?

sudo-bmitch: Looking back at the PR where we introduced them, I'd want to pull in other maintainers like @stevvooe to weigh in: https://github.com/opencontainers/distribution-spec/pull/111#issuecomment-996162334

extensions/_oci.md
```diff
@@ -66,6 +66,148 @@ Content-Type: application/json
...
+
+##### Query Parameters
+
+The following query parameters MAY be provided:
```
yosifkit, line 170: Perhaps there could also be a query parameter for "digest" such that a client could verify that a tag did point at the manifest when given an image reference like `image:tag@sha256...`.

It would return 404 if the digest isn't in the history but otherwise returning the expected JSON array of descriptor objects that includes the digest. I'm not sure if that'd be just the one descriptor (or two if deleted?) or perhaps up to `n`. And it should respect the other query parameters like `before` and so also 404 if the digest is not `before` the specified date.

extensions/_oci.md
yosifkit: Perhaps a registry could consider every update of a tag as a delete followed by a create and so the timestamp would be the same.

extensions/_oci.md
sudo-bmitch: Resolved with #612.

extensions/_oci.md
jcarter3: Ah, I think this was a mistake - it's meant to be histories for deleted _tags_ should continue to include the descriptors the tag pointed to when it existed. But it's already covered above

extensions/_oci.md
jcarter3: I added a comment for this above: 
```
If a pinned digest is provided with the tag, the registry MUST return a `404 Not Found` if the digest is not part of the tag's history.
```
This might cover it? Or we could send digest as a separate query param... maybe that's better?

yosifkit: 👍 I like pinned digest in the tag, no need for a query parameter when we already have the format. Oh, but I don't see the pinned digest format defined in `spec.md`? 🤔

I think it may need guidance on what they should return when passed a pinned digest?  Probably the only guarantee is that it must be in the list of `n` (barring `before` making it no longer in the list and thus 404).

extensions/_oci.md
```diff
@@ -66,6 +66,127 @@ Content-Type: application/json
...
+
+```HTTP
+GET /v2/<name>/_oci/tag-history/<tag>
+```
+
+`<name>` is the namespace of the repository, and `<tag>` is the name of the tag whose history is being queried.
+
+A successful request MUST return a `200 OK` response code.
+If the repository does not exist, the registry MUST return a `404 Not Found` response code.
+If the tag has no history or the registry does not implement this extension, the registry MUST return a `404 Not Found` response code.
```
yosifkit: Should this have a `MUST/SHOULD` respond with `UNSUPPORTED` in the error code for "registry does not implement this" case? Or perhaps some other ~recommended~ required error codes for the other 404 responses so that users of the API can know which it is. Or maybe the `n=0` trick is the correct way to figure out if history is supported?

---
Related:
> If the tag has no history [...] the registry MUST return a `404 Not Found`

> when `n` is zero, this endpoint MUST return an empty array

These seem conflicting. Perhaps a clarification is needed?

yosifkit: In other words, does `n=0` override the "repository does not exist" 404 or just the "no history" 404 or "registry does not implement this extension" 404 and what about the pinned digest 404?

jcarter3: Yes, those do some conflicting, good catch.

Currently most http routers on an unknown path would return 404 by default, hence the "return 404 if not supported". If we forced an UNSUPPORTED it'd require registries to explicitly update to say they aren't handling it, vs it being the default

jcarter3: I think the appropriate thing to do is return an empty response with 200 OK if there is no history - this seems to make more sense?

yosifkit: 🤔 Could it have no history? What does that mean? Does that mean nothing was ever tagged as this? Or just that the registry hasn't yet recorded history? But if it has at least the current tagged image and they support the history endpoint, it must have a minimum history of one, right? So, 404 is correct, the tag does not exist as an image, or it would have history. Returning an empty list for every tag that doesn't exist seems odd.

extensions/_oci.md
```diff
@@ -66,6 +66,133 @@ Content-Type: application/json
...
+Tag history MAY be retrieved with a standard `GET` as follows.
+
+```HTTP
+GET /v2/<name>/_oci/tag-history/<tag>
+```
+
+`<name>` is the namespace of the repository, and `<tag>` is the name of the tag whose history is being queried.
+
+A successful request MUST return a `200 OK` response code.
+If the repository does not exist, the registry MUST return a `404 Not Found` response code.
```
yosifkit: > If the repository does not exist, the registry MUST return a `404 Not Found` response code.

> When `n` is zero, this endpoint MUST return an empty array, useful for determining if history is available (200) or not (404) for the tag query.

I think someone could interpret that this should they return `200` and an empty list when `n=0` even if the repository does not exist.

sudo-bmitch (commented):
Adding our own annotations to the descriptors has me concerned. E.g. what if someone wants to pull up annotations in their own tag history API and there's a collision, perhaps even a collision with an annotation we haven't specified yet but add in a future release.

I'm playing around with my own implementation and have the following struct for each tag history entry:

```golang
type LayoutHistory struct {
	Time       time.Time  `json:"time"`
	Deleted    bool       `json:"deleted,omitempty"`
	Descriptor Descriptor `json:"descriptor"`
}
```

The descriptor ends up nested one extra layer in the responses, but we can now add other fields. E.g. I was considering if registries could add an audit log to each entry, identifying who (or what) was responsible for each entry.

---

Edit: A few more thoughts looking over this:

- I'd want to use the Link header for pagination. It's cleaner from the client side to know when more results are available.
- With the Link header, that gives flexibility for the server to reject some values of `n` for being too small or large, imposing it's own limits. Or the server can not support pagination and always return the full history.
- Do we have guidance for servers implementing garbage collection of history responses? If the server has a hard upper limit on the number of history entries per tag, that makes it easier to not paginate responses.
- When searching on a tag with a digest, I'd lean towards always returning a 200, and just having an empty list when no matches are found. Treat it like a filter on the history, where a tag could also have a particular digest assigned to it multiple times.
- Is pushing a new tag value treated as a simultaneous delete of the old value, or can that be implied?

extensions/_oci.md
```diff
@@ -66,6 +66,133 @@ Content-Type: application/json
...
+The following query parameters MAY be provided:
+
+- **`n`** *integer*, OPTIONAL
+
+  Specifies the maximum number of results to return.
+  The registry MAY return fewer results than requested if fewer historical entries exist.
+  When `n` is zero, this endpoint MUST return an empty array, useful for determining if history is available (200) or not (404) for the tag query.
+  When `n` is not specified, the registry MAY apply a default limit.
+
+- **`before`** *string (RFC 3339 timestamp)*, OPTIONAL
```
jonjohnsonjr, line 179: I'd love the opposite of this (`after` or `since`) to support the use case of "tailing" changes such that a client can cheaply check for updates to a tag to stay up to date. Some similar mechanisms would be `If-Modified-Since` headers or something like RSS. It would be great if we could lean on HTTP primitives for this.

tianon:
On today's call we talked about a lot of good reasons why it makes sense for this to be shaped like an OCI index (like #579 + referrers), and on the problem of "what if a registry 'pulls up' annotations?" we discussed having said registry ignore any image annotations that start with `org.opencontainers.distrubution.` (as explicitly "owned" by the registry), for future possible growth (so that clients can assume they're from the registry, not the image).

extensions/_oci.md
jcarter3: This has been updated. The only 404 is if the feature is not implemented. All other responses (missing/unknown repo/tag/digest) return an OCI Index with an empty manifest list.

extensions/_oci.md
jcarter3: Added a `since`

extensions/_oci.md
jcarter3: I've updated to include a `since`

extensions/_oci.md
```diff
@@ -66,6 +66,157 @@ Content-Type: application/json
...
+
+`<name>` is the namespace of the repository, and `<tag>` is the name of the tag whose history is being queried.
+
+A successful request MUST return a `200 OK` response code.
+If the registry does not implement this extension, the registry MUST return a `404 Not Found` response code.
+The registry MUST NOT return a `404 Not Found` response code for any other condition.
+If the repository does not exist or the tag has no history, the registry MUST return a `200 OK` response code with an empty manifest list.
+Tag history SHOULD remain queryable after the tag is deleted.
+Deleted manifests SHOULD NOT be removed from tag history.
+Tag history MAY be deleted after the repository is deleted.
```
michaelb990, line 89: from the example below, it seems like each registry operation (push or delete) will only create a single entry in the history. it seems a little ambiguous reading these requirements whether a tag moving between images creates 1 created event or 2 events (deleted/created).

can we add a requirement to clarify that a tag move, or a push to an existing tag, will only produce a single "created" event referencing the new image digest, assuming that's the direction we want to go?

extensions/_oci.md
```diff
@@ -66,6 +66,157 @@ Content-Type: application/json
...
+
+Upon success, the response body MUST be an OCI Index containing a list of descriptor objects in the following format:
```
michaelb990, line 91: i'm not sure about image indexes being used to represent content that may/may not exist...

the referrers API was a pretty natural use of the image index since it's designed to contain content that's actually in the registry and is going to be pulled, but this feels forced -- any thoughts on using more of a bare bones JSON list, similar to https://github.com/opencontainers/distribution-spec/pull/611?

potentially:
```
  GET /v2/team/web-app/_oci/tag-history/latest?n=3

  200 OK
  Content-Type: application/json

[
  {
    "event": "created",
    "timestamp": "2026-07-01T14:22:08.551204Z",
    "target": {
      "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
      "digest": "sha256:7d9f0e1c2b3a4958687776665554443332221110ffeeddccbbaa998877665544",
      "size": 3120
    }
  },
  {
    "event": "deleted",
    "timestamp": "2026-06-10T09:00:41.128870Z",
    "target": {
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "digest": "sha256:e3a1c07f9b2d4856a0f1c2e3d4b5a69788c7d6e5f4039281a0b1c2d3e4f50617",
      "size": 5871
    }
  },
  {
    "event": "created",
    "timestamp": "2026-05-04T03:02:01.043916Z",
    "target": {
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "digest": "sha256:e3a1c07f9b2d4856a0f1c2e3d4b5a69788c7d6e5f4039281a0b1c2d3e4f50617",
      "size": 5871
    }
  }
]
```

michaelb990 (commented):
a couple comments, but this seems like a cool idea!

for registries implementing this for existing repos, is there a way we should communicate to clients when we started tracking history? it seems like there's a difference between "this is the full history for a given tag" and "this is the history since 2026-08-15 for a given tag".

extensions/_oci.md
```diff
@@ -66,6 +66,157 @@ Content-Type: application/json
...
+```HTTP
+GET /v2/<name>/_oci/tag-history/<tag>
+```
+
+`<name>` is the namespace of the repository, and `<tag>` is the name of the tag whose history is being queried.
+
+A successful request MUST return a `200 OK` response code.
+If the registry does not implement this extension, the registry MUST return a `404 Not Found` response code.
+The registry MUST NOT return a `404 Not Found` response code for any other condition.
+If the repository does not exist or the tag has no history, the registry MUST return a `200 OK` response code with an empty manifest list.
```
northtyphoon, line 86: >If the repository does not exist or the tag has no history

the statement is a bit ambiguous, a tag must have a history, at least `created` if it ever exits. It's better to reword to something like `if the repository or tag never exits`

extensions/_oci.md
```diff
@@ -66,6 +66,157 @@ Content-Type: application/json
...
+  The digest of the manifest, in the form `<algorithm>:<encoded>`.
+
+- **`annotations`** *map of strings*, REQUIRED
+
+  Annotations associated with the historical tag entry. Registries MAY include annotations from the underlying manifest or index descriptor.
+- When doing so, any existing annotations with keys beginning with `org.opencontainers.distribution` MUST be ignored because this prefix is reserved for annotations generated by the registry.
+
+  The annotations MUST include both of the following:
+
+  - **`org.opencontainers.distribution.tag.timestamp`** *string*
```
northtyphoon, line 160: `timestamp` is a universal concept, suggest remove `tag` from the annotation name

extensions/_oci.md
northtyphoon: I would think the history can be more compact for repush case. Actually two created events are better repressing what acutally happened compared to created/deleted/created.

extensions/_oci.md
jcarter3: Yes, I can add clarifying language there - the intent was to do a single "create" event when a tag moves, and "deleted" events only apply if it is specifically removed and no longer referencable in the registry.

extensions/_oci.md
jcarter3: We debated this a couple times - ultimately opted to stick with the index approach rather than having to define a new schema/model that everyone needs to copy/import. The data format isn't _ideal_ but at least it is known. Happy to revisit if more people challenge it but that was the consensus before.

tianon: Yeah, IMO the benefits to making all our APIs use our existing data types, even if it's a stretch, outweigh the downsides. :sweat_smile:

sudo-bmitch: I'm one of the few that would be happy to see a new media type for this, but was out voted last time it came up. That opinion is a weak one, so no big deal if this moves forward using the OCI Index.

PR #611, opened 2026-05-07
Update spec to include repository endpoint

jcarter3:
Throwing out a proposal to replace/enhance _catalog

### DIFF
--- a/extensions/_oci.md
+++ b/extensions/_oci.md
@@ -66,6 +66,105 @@ Content-Type: application/json
 
     Enumeration of the endpoints provided on this registry (as not all "OPTIONAL" endpoints may be present in all registries)
 
+### Component: `repositories`
+
+This component is for endpoints relating to repository listing operations.
+
+This endpoint returns repositories under a namespace path, in lexical (i.e. case-insensitive alphanumeric order) or "ASCIIbetical" ([Go's `sort.Strings`](https://pkg.go.dev/sort#Strings)) order.
+
+Repository listing MAY be retrieved with a standard `GET` as follows.
+
+```HTTP
+GET /v2/<name>/_oci/repositories
+```
+
+Registry implementations MAY allow `<name>` to be empty in order to list all repositories available on the registry.
+When `<name>` is empty, the path is:
+
+```HTTP
+GET /v2/_oci/repositories
+```
+
+A registry that does not support listing repositories with an empty `<name>` MUST return a `405 Method Not Allowed` response code for this request.
+
+When non-empty, `<name>` is a `/` delimited repository namespace path.
+For this endpoint, a non-empty `<name>` is used as a namespace segment prefix and MUST include one or more complete namespace segments.
+For requests with a non-empty `<name>`, returned repositories MUST have names that are equal to `<name>` or that begin with `<name>/`.
+Registries MUST NOT treat a non-empty `<name>` as a partial namespace segment prefix.
+For example, given repositories `foo/bar`, `foo/baz`, and `bar/baz`, a request to `/v2/foo/_oci/repositories` returns `foo/bar` and `foo/baz`.
+
+A successful request MUST return a `200 OK` response code.
+If the namespace path is invalid or the registry cannot list repositories under it, the registry MUST return an appropriate error response.
+The list of repositories MAY be empty if there are no repositories under the namespace path.
+Registry implementations MAY limit the repositories returned based on implementation policy, access control, or other registry-specific behavior.
+The presence of a repository in the response only indicates that the registry may provide access to that repository at the time of the request.
+The absence of a repository from the response does not indicate that the repository does not exist.
+
+Upon success, the response body MUST be a JSON array of repository objects in the following format:
+
+```HTTP
+200 OK
+Content-Length: <length>
+Content-Type: application/json
+
+[
+    {
+        "name": "foo/bar"
+    },
+    {
+        "name": "foo/baz"
+    }
+]
+```
+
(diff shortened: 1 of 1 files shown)


extensions/_oci.md
```diff
@@ -66,6 +66,105 @@ Content-Type: application/json
...
+### Component: `repositories`
+
+This component is for endpoints relating to repository listing operations.
+
+This endpoint returns repositories under a namespace path, in lexical (i.e. case-insensitive alphanumeric order) or "ASCIIbetical" ([Go's `sort.Strings`](https://pkg.go.dev/sort#Strings)) order.
+
+Repository listing MAY be retrieved with a standard `GET` as follows.
+
+```HTTP
+GET /v2/<name>/_oci/repositories
```
sudo-bmitch, line 78: To follow the `_extension/module/component` syntax, perhaps:

```suggestion
GET /v2/<name>/_oci/repository/list
```

extensions/_oci.md
```diff
@@ -66,6 +66,105 @@ Content-Type: application/json
...
+```
+
+Registry implementations MAY allow `<name>` to be empty in order to list all repositories available on the registry.
+When `<name>` is empty, the path is:
+
+```HTTP
+GET /v2/_oci/repositories
```
sudo-bmitch, line 85:
```suggestion
GET /v2/_oci/repository/list
```

extensions/_oci.md
```diff
@@ -66,6 +66,105 @@ Content-Type: application/json
...
+```
+
+A registry that does not support listing repositories with an empty `<name>` MUST return a `405 Method Not Allowed` response code for this request.
```
sudo-bmitch, line 88: Shouldn't it be 404, same as a registry that doesn't implement the extension? I'd also consider "SHOULD" since implementations are really good at throwing other 4xx codes, along with the occasional 5xx. We'd like registries to be consistent, but clients should be prepared for a wild west.

extensions/_oci.md
```diff
@@ -66,6 +66,105 @@ Content-Type: application/json
...
+
+When non-empty, `<name>` is a `/` delimited repository namespace path.
```
sudo-bmitch, line 90: Perhaps we should say `<prefix>` or some other term since `<name>` is used throughout the spec to refer to the full namespace path.

extensions/_oci.md
```diff
@@ -66,6 +66,105 @@ Content-Type: application/json
...
+For this endpoint, a non-empty `<name>` is used as a namespace segment prefix and MUST include one or more complete namespace segments.
+For requests with a non-empty `<name>`, returned repositories MUST have names that are equal to `<name>` or that begin with `<name>/`.
+Registries MUST NOT treat a non-empty `<name>` as a partial namespace segment prefix.
+For example, given repositories `foo/bar`, `foo/baz`, and `bar/baz`, a request to `/v2/foo/_oci/repositories` returns `foo/bar` and `foo/baz`.
```
sudo-bmitch, line 94: To make it clear the segments must be complete, I'd include a matching partial segment:

```suggestion
For example, given repositories `foo/bar`, `foo/baz`, and `foobar/baz`, a request to `/v2/foo/_oci/repositories` returns `foo/bar` and `foo/baz`.
```

extensions/_oci.md
```diff
@@ -66,6 +66,105 @@ Content-Type: application/json
...
+
+A successful request MUST return a `200 OK` response code.
+If the namespace path is invalid or the registry cannot list repositories under it, the registry MUST return an appropriate error response.
```
sudo-bmitch, line 97: We should list the expected error response codes. I'm expecting 400, 401, 403, and 404 to all be expected errors to cover invalid prefix, not found prefix, unauthorized, and extension not implemented.

extensions/_oci.md
```diff
@@ -66,6 +66,105 @@ Content-Type: application/json
...
+
+Each repository object in the response MUST include the following properties:
+
+- **`name`** *string*, OPTIONAL
+
+  The full repository name.
+  The value MUST match the repository name syntax defined for `<name>` in this specification.
+  The value MUST be equal to the requested namespace path or have the requested namespace path as a complete namespace segment prefix.
+
+Repository objects MAY include additional properties.
```
sudo-bmitch, line 128:
```suggestion
Repository objects MAY include additional properties.
Field order and whitespace may vary by implementation.
```

extensions/_oci.md
```diff
@@ -66,6 +66,105 @@ Content-Type: application/json
...
+
+- **`last`** *string*, OPTIONAL
+
+  Specifies the `name` value of the last repository object received by the client.
+  When provided, the response MUST begin non-inclusively after `last` in the ordered repository list.
+  The `last` value MUST NOT be a numerical index.
+  When using the `last` query parameter, the `n` parameter is OPTIONAL.
+
+A `Link` header MAY be included in the response when additional repositories are available.
+If included, the `Link` header MUST be set according to [RFC5988](https://www.rfc-editor.org/rfc/rfc5988.html) with the Relation Type `rel="next"`.
```
sudo-bmitch, line 150: I'm sure the pagination was pulled from the tag listing. I think there's debate on whether the `Link` header or the `last` parameter should be used, and for historical reasons we ended up needing to document/support both. For a green field API, we have an opportunity to pick a single pagination option, and I'd probably lean towards the `Link` header for that. Then we could remove the `n` and `last` parameters.

sudo-bmitch (changes requested):
Just a first pass. As with the tag listing, I'm going to want to spend some time implementing it and see a conformance test before approving.

PR #614, opened 2026-06-19
Fix: Upload artifacts in conformance test

sudo-bmitch:
This fixes the generation of artifacts in the conformance results. When the variable isn't exported, the default "results" directory is used for outputting the results.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -75,7 +75,7 @@ runs:
       run: |
         set -x
         set +e
-        OCI_RESULTS_DIR=${OCI_RESULTS_DIR:-.}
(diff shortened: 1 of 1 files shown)


sudo-bmitch:
The difference can be seen in these actions:

- Before: https://github.com/olareg/olareg/actions/runs/27836089590
- After: https://github.com/olareg/olareg/actions/runs/27836237711

mikebrow (approved):
LGTM

tianon (approved):

rchincha (approved):

dmcgowan (approved):

PR #615, opened 2026-06-27
OCI copyright headers

sudo-bmitch:
This aligns with the TOB guidelines in https://github.com/opencontainers/tob/pull/159

### DIFF
--- a/.tool/check-license
+++ b/.tool/check-license
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 # Copyright the Open Container Initiative Contributors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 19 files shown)


.tool/check-license
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ret=0
+
+for file in $(find . -type f -iname '*.go' ! -path './vendor/*'); do
```
mikebrow, line 23: fair to exclude vendor.. but we don't have any ..

mikebrow (approved):
LGTM

.tool/check-license
sudo-bmitch: This was copied from [image-spec](https://github.com/opencontainers/image-spec/blob/main/.tool/check-license), which I assume was handling the case the where someone vendored the deps to do some offline dev work.

dmcgowan (approved):

Issue #610, opened 2026-05-07
Provide some guidance for `<tag>@<digest>` references


Silvanoc:
It can be observed lately that more and more clients are supporting the use of `<tag>@<digest>` references for OCI-artifacts in general on their user interface. This naming format [conflicts with the OCI-specification](https://github.com/opencontainers/distribution-spec/blob/c33cde3562c5753cb9dcd15d5afdeee9be4ef553/spec.md?plain=1#L153-L154), which clearly states that "_`<reference>` MUST be either (a) the digest of the manifest or (b) a tag. The `<reference>` MUST NOT be in any other format._".

It's clear that the user-client interface is not bound to the OCI-specification. Nevertheless, it would be meaningful to give some guidance on this naming format to ensure consistency across the different levels in the ecosystem.

I see two options:

1. Keeping the OCI-specification as it is, but give some guidance on how to deal with it on other levels in the ecosystem.
2. Extending the specification to make `<tag>@<digest>` also valid, specifying the expected behavior (probably ignoring the <tag> part).

@sudo-bmitch apparently _regctl_ supports this format and you are a maintainer of both _regctl_ and this specification. What's your opinion on this topic?

sudo-bmitch:
This spec defines the registry API, where no conflict exists (except for overloading the reference term which I feel the spec needs to pick a new term).

There is a proposal to create a working group to standardize the reference strings used to point to images, but we don't even have enough people interested in owning that effort, along with disagreement on whether OCI should standardize that: https://github.com/opencontainers/tob/pull/114

Until then, the reference strings are defined by individual projects. Distribution, regclient, and skopeo all have separate implementations with various incompatibilities.

- https://github.com/distribution/reference
- https://github.com/regclient/regclient/tree/main/types/ref and https://regclient.org/usage/#image-references
- https://github.com/containers/image/blob/main/docs/containers-transports.5.md

sudo-bmitch:
I'm going to close this out since I don't believe this is a distribution spec issue (the reference string used by applications != the path segment in the HTTP API, these are two separate things).

PR #616, opened 2026-06-30
Remove "reference" from spec

### LINKED ISSUE #216: Improving usage of `reference` terminology

sudo-bmitch:
There were two places the term "reference" was used:

1. As the "tag or digest" parameter in the manifest endpoints. This was changed to "tag-or-digest".
2. As the returned location to upload blobs in our endpoints table. This was changed to "blob-push-location", and then usage of "location" earlier in the spec were updated for consistency.

This change removes confusion with the "reference" term in the OCI ecosystem where it is a string that identifies an image, including the registry, repository, and identifier. For context, see https://github.com/distribution/reference.

Fixes #216.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -141,11 +141,11 @@ Typically, the first step in pulling an object is to retrieve the manifest. Howe
 #### Pulling manifests
 
 To pull a manifest, perform a `GET` request to a URL in the following form:
-`/v2/<name>/manifests/<reference>` <sup>[end-3](#endpoints)</sup>
+`/v2/<name>/manifests/<tag-or-digest>` <sup>[end-3](#endpoints)</sup>
 
 `<name>` refers to the namespace of the repository.
-`<reference>` MUST be either (a) the digest of the manifest or (b) a tag.
-The `<reference>` MUST NOT be in any other format.
+`<tag-or-digest>` MUST be either (a) the digest of the manifest or (b) a tag.
+The `<tag-or-digest>` MUST NOT be in any other format.
 Throughout this document, `<name>` MUST match the following regular expression:
 
 `[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*`
@@ -155,7 +155,7 @@ Many clients impose a limit of 255 characters on the length of the concatenation
 If the registry name is `registry.example.org:5000`, those clients would be limited to a `<name>` of 229 characters (255 minus 25 for the registry hostname and port and minus 1 for a `/` separator).
 For compatibility with those clients, registries should avoid values of `<name>` that would cause this limit to be exceeded.
 
-Throughout this document, `<reference>` as a tag MUST be at most 128 characters in length and MUST match the following regular expression:
+Throughout this document, `<tag-or-digest>` as a tag MUST be at most 128 characters in length and MUST match the following regular expression:
 
 `[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}`
 
@@ -174,7 +174,7 @@ The `Docker-Content-Digest` header, if present on the response, returns the dige
 If the digest does differ, it MAY be the case that the hashing algorithms used do not match.
 See [Content Digests](https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#digests) <sup>[apdx-3](#appendix)</sup> for information on how to detect the hashing algorithm in use.
 Most clients MAY ignore the value, but if it is used, the client MUST verify the value matches the returned manifest.
-If the `<reference>` part of a manifest request is a digest, clients SHOULD verify the returned manifest matches this digest.
+If the `<tag-or-digest>` part of a manifest request is a digest, clients SHOULD verify the returned manifest matches this digest.
 
 If the manifest is not found in the repository, the response code MUST be `404 Not Found`.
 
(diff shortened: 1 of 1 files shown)


spec.md
```diff
@@ -141,11 +141,11 @@ Typically, the first step in pulling an object is to retrieve the manifest. Howe
 #### Pulling manifests
 
 To pull a manifest, perform a `GET` request to a URL in the following form:
-`/v2/<name>/manifests/<reference>` <sup>[end-3](#endpoints)</sup>
+`/v2/<name>/manifests/<identifier>` <sup>[end-3](#endpoints)</sup>
```
mikebrow:
```suggestion
`/v2/<name>/manifests/<manifest identifier>` <sup>[end-3](#endpoints)</sup>
```

spec.md
```diff
@@ -141,11 +141,11 @@ Typically, the first step in pulling an object is to retrieve the manifest. Howe
...
 
 `<name>` refers to the namespace of the repository.
-`<reference>` MUST be either (a) the digest of the manifest or (b) a tag.
-The `<reference>` MUST NOT be in any other format.
+`<identifier>` MUST be either (a) the digest of the manifest or (b) a tag.
```
mikebrow:
```suggestion
`<manifest identifier>` (previously defined simply as reference) MUST be either (a) the digest of the manifest or (b) a tag.
```

spec.md
```diff
@@ -141,11 +141,11 @@ Typically, the first step in pulling an object is to retrieve the manifest. Howe
...
+The `<identifier>` MUST NOT be in any other format.
```
mikebrow:
```suggestion
The `<manifest identifier>` MUST NOT be in any other format.
```

mikebrow (commented):
thoughts?

spec.md
sudo-bmitch: I'd want to have this discussion in #216 since the `<identifier>` term was proposed and agreed to there.

spec.md
sudo-bmitch: Would indicating this term was previously known as `<reference>` add clarity or continue the confusion with the other definitions of reference?

mikebrow: to be honest identifier is a much more over used term.. the clarity is to the people who are greping/googling for "identifier|manifest identifier" and finding nothing that relates to distribution services in their registry/client code bases?

mikebrow: for example see CRI api struggles with image id/reference naming conventions and getting different use patterns based on the various assumptions for the overloaded terms.

```
	// Reference to the unique identifier of the image on the node, as
	// returned in the image and runtime service apis.
	//
	// When referring to the same image, the container runtime MUST always return
	// the same value for:
	//   - Image.id
	//   - Container.image_id
	//   - ContainerStatus.image_id
	//   - PullImageResponse.image_ref
	//
	// Note: this field has a stricter meaning starting with v1.36.
	// It used to be image ID OR digest, which are different values,
	// and would not necessarily match other ID fields in the Container,
	// ContainerStatus, and Image messages
	ImageRef      string `protobuf:"bytes,1,opt,name=image_ref,json=imageRef,proto3" json:"image_ref,omitempty"`

```

sudo-bmitch: I'm not too concerned about overused (we also have `<name>` all over the spec). But I'm concerned of differently used, where "reference" is used to mean something different in all those OCI implementing code bases. Grepping for that term is likely to send them down paths of confusion like #610.

mikebrow: nod.. the example I showed also had a lot of confusion with cri-o/containerd initially using different ID types in their respective pull response and image status apis.  pain

mikebrow: the name issue is related to the namespace exclusion policy.. :)

sudo-bmitch:
This has been updated with the discussion in #216 to use `<tag-or-digest>` for the previous usage of `<reference>` in the manifest endpoints.

mikebrow (approved):
LGTM

spec.md
```diff
@@ -141,11 +141,11 @@ Typically, the first step in pulling an object is to retrieve the manifest. Howe
 #### Pulling manifests
 
 To pull a manifest, perform a `GET` request to a URL in the following form:
-`/v2/<name>/manifests/<reference>` <sup>[end-3](#endpoints)</sup>
+`/v2/<name>/manifests/<tag-or-digest>` <sup>[end-3](#endpoints)</sup>
 
 `<name>` refers to the namespace of the repository.
-`<reference>` MUST be either (a) the digest of the manifest or (b) a tag.
-The `<reference>` MUST NOT be in any other format.
+`<tag-or-digest>` MUST be either (a) the digest of the manifest or (b) a tag.
```
tianon, line 147: I kind of love that these are in the opposite order; "`<tag-or-digest>` must be .. a digest or a tag" 😂

tianon (approved):

dmcgowan (approved):

PR #617, opened 2026-06-30
Fix: Set conformance commit in GitHub Action

sudo-bmitch:
The syntax for setting a variable in main is different from setting it in packages. This fixes how we run it within GitHub Actions (where there isn't a .git directory).


dmcgowan (approved):

mikebrow (approved):
LGTM

PR #619, opened 2026-07-16
Clarify blob-push-location

sudo-bmitch:
This comes out of today's meeting, where we didn't want to lose the historical `/v2/<name>/blobs/uploads/<unique-id>` value of the blob push location.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -253,12 +253,16 @@ Upon success, the response MUST have a code of `202 Accepted`, and MUST include 
 Location: <blob-push-location>
 ```
 
-The `<blob-push-location>` MUST contain a UUID representing a unique session ID for the upload to follow.
-The `<blob-push-location>` does not necessarily need to be provided by the registry itself.
+Registries MUST support concurrent uploads from one or more clients, which is commonly implemented with a unique `<blob-push-location>` for each blob upload.
+
+The `<blob-push-location>` MAY be provided by a server other than the registry itself.
 In fact, offloading to another server can be a [better strategy](https://www.backblaze.com/blog/design-thinking-b2-apis-the-hidden-costs-of-s3-compatibility/).
 
-Optionally, the location MAY be absolute (containing the protocol and/or hostname), or it MAY be relative (containing just the URL path).
+The `<blob-push-location>` MAY be absolute (containing the protocol and/or hostname), or it MAY be relative (containing just the URL path).
 For more information, see [RFC 7231 (section 7.1.2)](https://www.rfc-editor.org/rfc/rfc7231#section-7.1.2).
+
+_Implementers note:_
+Registries commonly set `<blob-push-location>` to `/v2/<name>/blobs/uploads/<unique-id>`, where `<unique-id>` is a unique identifier.
 
 Once the `<blob-push-location>` has been obtained, perform the upload proper by making a `PUT` request to the following URL path, and with the following headers and body:
 
@@ -272,8 +276,8 @@ Content-Type: application/octet-stream
 ```
 
 The `<blob-push-location>` MAY contain critical query parameters.
-Additionally, it SHOULD match exactly the `<blob-push-location>` obtained from the `POST` request.
-It SHOULD NOT be assembled manually by clients except where absolute/relative conversion is necessary.
+Additionally, it MUST match exactly the `<blob-push-location>` obtained from the `POST` request.
+It MUST NOT be assembled manually by clients except where absolute/relative conversion is necessary.
 
 Here, `<digest>` is the digest of the blob being uploaded, and `<length>` is its size in bytes.
 


spec.md
```diff
@@ -253,11 +253,12 @@ Upon success, the response MUST have a code of `202 Accepted`, and MUST include
 Location: <blob-push-location>
 ```
 
-The `<blob-push-location>` MUST contain a UUID representing a unique session ID for the upload to follow.
-The `<blob-push-location>` does not necessarily need to be provided by the registry itself.
+The `<blob-push-location>` MUST contain a unique ID for the upload to follow.
+The `<blob-push-location>` SHOULD be `/v2/<name>/blobs/uploads/<unique-id>`, where `<unique-id>` is a unique ID.
```
sudo-bmitch: Today's meeting wanted a SHOULD here. Given the rest of this section, I think an "implementers note" or MAY would make more sense.

spec.md
```diff
@@ -253,11 +253,12 @@ Upon success, the response MUST have a code of `202 Accepted`, and MUST include
...
+The `<blob-push-location>` MAY be provided by the a server other than registry itself.
```
dmcgowan: Clarify the sentence more

mikebrow:
```suggestion
The `<blob-push-location>` MAY be provided by a server other than the registry itself.
```

sudo-bmitch: Thanks Mike, fixed that. This was rewording the old line 257 a bit, and goes with the next line that talks about offloading the upload processing to another server.

spec.md
rchincha: "unique ID"

Scope of "unique"? within repo? registry? global?

spec.md
```diff
@@ -272,8 +276,8 @@ Content-Type: application/octet-stream
 ```
 
 The `<blob-push-location>` MAY contain critical query parameters.
```
sudo-bmitch, line 278: Is there an RFC or similar upstream spec we can point to that clarifies how to append the `digest=<digest>` query parameter to a URI that already contains query parameters, including escaping of any special characters.

tianon: Man.  That's a rabbit hole.  The short answer is "no" there is _not_ a canonical RFC for this format.  It's basically as old as the world wide web itself.

The closest thing to an "RFC" is https://datatracker.ietf.org/doc/html/rfc1866#section-8.2.1 (yes, HTML 2.0).

There is a _very_ slight *nod* to the format in https://datatracker.ietf.org/doc/html/rfc3986#section-3.4, which is the closest thing to it being canonical for URLs (`query components are often used to carry identifying information in the form of "key=value" pairs and one frequently used value is a reference to another URI`).

*Officially*, I think the format is technically "defined" by the `mediaType`, `application/x-www-form-urlencoded`: https://www.iana.org/assignments/media-types/application/x-www-form-urlencoded (official documentation at https://url.spec.whatwg.org/#application/x-www-form-urlencoded) -- however, note that WHATWG is a *living* spec, so while the format itself is solid, the references to it and the descriptions of it might change/move over time.

sudo-bmitch: Welp, we can leave this ambiguous, and kick this can of worms down the road, if that's easiest.

tianon: It's a little cheeky, but we could link to my comment here if you wanted somewhere to link without reproducing _all_ that text in the spec itself. :joy:

sudo-bmitch: I think the change I want to make should be a separate PR. Where we say:

> The `<blob-push-location>` MAY contain critical query parameters.
> Additionally, it MUST match exactly the `<blob-push-location>` obtained from the `POST` request.
> It MUST NOT be assembled manually by clients except where absolute/relative conversion is necessary.

I'm leaning towards:

> The `<blob-push-location>` MAY contain critical query parameters.
> Clients MUST assemble the URI from the `<blob-push-location>` following relative resolution ([RFC 3986 section 5.2](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2)) and encoding the `digest=<digest>` query parameter ([RFC 3986 section 3.4](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4)) using [`application/x-www-form-urlencoded` syntax](https://url.spec.whatwg.org/#application/x-www-form-urlencoded).

Hopefully that better captures what servers and clients are already doing today. I really hope no one is blindly appending a `?digest=<digest>` string to the received URL.

dmcgowan (dismissed):

spec.md
```diff
@@ -253,13 +253,17 @@ Upon success, the response MUST have a code of `202 Accepted`, and MUST include
 Location: <blob-push-location>
 ```
 
-The `<blob-push-location>` MUST contain a UUID representing a unique session ID for the upload to follow.
-The `<blob-push-location>` does not necessarily need to be provided by the registry itself.
+The `<blob-push-location>` MUST support concurrent uploads from one or more clients.
```
tianon: Does this mean multiple clients using that _same_ blob push location value? (that's how it reads to me, and that feels wrong but maybe I'm just sluggish today)

sudo-bmitch: Help me with some better phrasing. It's supposed to mean one or more clients, pushing multiple blobs, would not interfere with each other. It's focused on the outcome and less prescriptive than "thy shall use uuids".

tianon: I recall we were trying to get away from the word "unique" but I don't recall the exact arguments we had:

```suggestion
The `<blob-push-location>` MUST be unique to *this* upload somehow (UUID, etc).
```

I *think* what was problematic before was that we specifically said "unique ID" which felt too prescriptive on the specific method of uniqueness.

sudo-bmitch: I took a shot at rephrasing it to clarify we expect the registry and not a specific URL to support concurrent uploads.

tianon (approved):

dmcgowan (approved):

PR #622, opened 2026-08-28
Add `DisableCompression: true` to our conformance `Transport`

tianon:
Technically, the spec *does* support transport compression by never making `Content-Length` a `MUST` on a `GET`, but the conformance tests currently don't agree with that, so running them against a server that *does* implement transport compression has Go's `net/http` by-default requesting transport compression and transparently stripping it on response bodies, but then having to strip `ContentLength` (because it's over the compressed data).

As a result, over 200 tests fail on such a server. 😂

On the flip side, we haven't really seen this much in practice because Go's *server* does not implement transparent transport compression like the client does, so this only shows up if a server explicitly implements it.

This is intended to be a temporary band-aid, by disabling Go's transparent transport compression handling, which fixes the test failures on a server which actually implements transport compression.

- see also https://github.com/cue-labs/oci/issues/38 (where I've bumped into this problem before)

### DIFF
--- a/conformance/run.go
+++ b/conformance/run.go
@@ -65,9 +65,20 @@ func runnerNew(c config) (*runner, error) {
 	if c.LoginUser != "" && c.LoginPass != "" {
 		apiOpts = append(apiOpts, apiWithAuth(c.LoginUser, c.LoginPass, c.CacheAuth))
 	}
+
+	// TODO <tianon> ideally we wouldn't just blatantly DisableCompression here, but we currently (2026-08-27) have a bunch of tests that assume Content-Length is set that Go's transparent handling of transport compression makes fail if a server happens to implement transport compression
+	client := &http.Client{}
+	if t, ok := http.DefaultTransport.(*http.Transport); ok {
+		transport := t.Clone()
+		transport.DisableCompression = true
+		client.Transport = transport
+	} else {
+		return nil, fmt.Errorf("changing the DefaultTransport is not supported")
+	}
+
 	r := runner{
 		Config:  c,
-		API:     apiNew(http.DefaultClient, apiOpts...),
+		API:     apiNew(client, apiOpts...),
 		State:   stateNew(),
 		Results: resultsNew(testName, nil),
 		Log:     slog.New(slog.NewTextHandler(c.LogWriter, &slog.HandlerOptions{Level: lvl})),


tianon:
TODO:

- survey clients to see if https://github.com/cue-labs/oci/issues/38 is an outlier or the norm (and thus whether transport compression on a server will break _actual_ clients in practice, not just the conformance tests)

- depending on that survey, decide what to do in the spec / conformance (fix the conformance test assumptions, add a warning to the spec about clients, explicitly disallow transport compression, etc)

conformance/run.go
```diff
@@ -67,7 +67,7 @@ func runnerNew(c config) (*runner, error) {
 	}
 	r := runner{
 		Config:  c,
-		API:     apiNew(http.DefaultClient, apiOpts...),
+		API:     apiNew(&http.Client{Transport: &http.Transport{DisableCompression: true}}, apiOpts...), // TODO <tianon> ideally we wouldn't just blatantly DisableCompression here, but we currently (2026-08-27) have a bunch of tests that assume Content-Length is set that Go's transparent handling of transport compression makes fail if a server happens to implement transport compression
```
sudo-bmitch: I think we need to clone the `http.DefaultTransport` to ensure we get the proxy and timeout settings described at https://pkg.go.dev/net/http@go1.27.0#DefaultTransport:

```
client := &http.Client{}
if t, ok := http.DefaultTransport.(*http.Transport); ok {
  client.Transport = t.Clone()
  client.Transport.DisableCompression = true
} else {
  return nil, fmt.Errorf("Changing the DefaultTransport is not supported")
}
...
apiNew(client, apiOpts...)
```

sudo-bmitch (changes requested):

conformance/run.go
tianon: Had to make some tiny tweaks to appease Go, but pushed :+1:

sudo-bmitch (dismissed):

sudo-bmitch:
Doh, staticcheck is unhappy with that error message, which I should have known better to not capitalize.

tianon:
> TODO:
> 
> * survey clients to see if [ociclient: `Content-Length` "SHOULD" be present, but isn't guaranteed cue-labs/oci#38](https://github.com/cue-labs/oci/issues/38) is an outlier or the norm (and thus whether transport compression on a server will break _actual_ clients in practice, not just the conformance tests)
> * depending on that survey, decide what to do in the spec / conformance (fix the conformance test assumptions, add a warning to the spec about clients, explicitly disallow transport compression, etc)

I did some paperclip counting on this:

---

**Servers (5 implementations checked: `distribution`, `zot`, `olareg`, `cue-labs/oci`, `google/go-containerregistry`'s test registry):** none of them apply `Content-Encoding` compression to distribution API responses. No compression middleware anywhere, and the one documented reverse-proxy recipe (`distribution`'s nginx doc) doesn't enable it either. So none of the reference servers in this corpus would ever trigger the scenario this TODO is worried about.

**Clients (9 implementations checked):** of the 8 that actually talk to a registry over HTTP, 7 already tolerate a missing/unknown `Content-Length` on manifest/blob GET responses without erroring:

- `containerd`, `distribution` (client side) — both explicitly set their own `Accept-Encoding` header, which sidesteps Go's transparent-gzip-decompression behavior (and the `Content-Length`-stripping that comes with it) entirely, rather than disabling compression negotiation
- `moby` (via containerd), `google/go-containerregistry`, `skopeo` (containers/image) — read bodies via `io.ReadAll`/EOF-driven reads and never hard-require `ContentLength >= 0`
- `regclient` — tolerant, with a minor caveat: it loses an early-truncation-retry optimization when length is unknown, but doesn't fail
- `docker-cli` and `umoci` don't do registry HTTP I/O themselves (delegate to `dockerd`/don't talk to registries), so not applicable

One exception: **`cue-labs/oci`** does hard-require `Content-Length` in its manifest/blob read path (`descriptorFromResponse()` in `ociregistry/ociclient/client.go` errors `"unknown content length"` when `resp.ContentLength < 0`) — this looks like a genuine bug independent of this PR, worth reporting upstream separately.

Given that, this looks like a synthetic edge case for the conformance suite rather than real-world client/server behavior — I don't think any of these implementations would be exercised by a compressing server in practice. If it's worth avoiding `DisableCompression` as a permanent fixture, the pattern used by `containerd`/`distribution`'s client (explicitly setting `Accept-Encoding` on the test transport) might be a closer match to how real clients avoid this than disabling compression negotiation outright — but `DisableCompression: true` as a stopgap seems reasonable given how rare the underlying scenario is.

tianon:
So to summarize, it looks like https://github.com/cue-labs/oci/issues/38 really _is_ an outlier, not the norm, and we could very reasonably consider both implementing and testing transport compression.

tianon:
That cagey language around regclient was bothering me (given it's relevant to Brandon directly), so I pressed it to dig deeper:

---

Small correction/refinement to the regclient entry in my earlier survey: I initially described it as "tolerant, with a caveat" (losing an early-truncation-retry safety net when `Content-Length` is unknown). On closer look at [`request.Read()`](https://github.com/regclient/regclient/blob/1a4d357a3a6df1d4d4164bb1aa110fe0259a6c30/internal/retryable/retryable.go#L603-L648), that's overstated — the "retry" in question is actually dead code (commented out, with a literal [`// TODO: handle early EOF ... with a retry`](https://github.com/regclient/regclient/blob/1a4d357a3a6df1d4d4164bb1aa110fe0259a6c30/internal/retryable/retryable.go#L621) right above it), so today an early EOF gets the same outcome — a plain `io.EOF` — whether or not `Content-Length` was known. The only thing that differs is a debug log line that fires before that already-inert stub returns.

The actual correctness backstop — digest verification on read — doesn't depend on `Content-Length` at all. It lives one layer up, in [`types/blob.BReader`](https://github.com/regclient/regclient/blob/1a4d357a3a6df1d4d4164bb1aa110fe0259a6c30/types/blob/reader.go#L103-L130), which tees the body through its own digester and checks it against the expected digest (from the caller's descriptor, or the `Docker-Content-Digest` response header) on [`Close`](https://github.com/regclient/regclient/blob/1a4d357a3a6df1d4d4164bb1aa110fe0259a6c30/types/blob/reader.go#L86-L95)/EOF, regardless of whether the length was known in advance.

So regclient should just move to the "tolerant" column outright: no dependency on `Content-Length`, and no loss of any currently-functioning safety net when it's absent.

sudo-bmitch (approved):

tianon:
If you'd like to test a client against a registry that enables transport compression, https://github.com/distribution/distribution/commit/86f3f1ffec971bbff32cd04549e49f73ce24dbd0 slash https://github.com/distribution/distribution/tree/86f3f1ffec971bbff32cd04549e49f73ce24dbd0 has you covered.

sudo-bmitch:
> Small correction/refinement to the regclient entry in my earlier survey... On closer look at [`request.Read()`](https://github.com/regclient/regclient/blob/1a4d357a3a6df1d4d4164bb1aa110fe0259a6c30/internal/retryable/retryable.go#L603-L648), that's overstated — the "retry" in question is actually dead code...

Not just dead code, dead code in a dead package, that whole thing will get pruned in a not too distant release:

```golang
//go:build legacy
// +build legacy

// Package retryable is a legacy package, functionality has been moved to reghttp
```

jdolitsky (approved):

PR #624, opened 2026-09-01
Implement transport compression in the conformance tests

tianon:
This is conservative -- it makes `Content-Length` optional on most tests, but *only* if we know we're dealing with transport compression.

It then also explicitly tests transport compression, but does not fail for it missing: it only fails if it's implemented incorrectly (otherwise the tests are Skipped).

- follow-up to https://github.com/opencontainers/distribution-spec/pull/622

Assisted-By: "claude my eyes right out"

### DIFF
--- a/conformance/api.go
+++ b/conformance/api.go
@@ -227,7 +227,7 @@ func (a *api) BlobGetExistsFull(registry, repo string, dig digest.Digest, td *te
 		apiReturnResponse(&resp),
 	)
 	if val, ok := td.blobs[dig]; ok && (len(val) > 0 || dig == emptyDigest) {
-		opts = append(opts, apiExpectBody(val), apiExpectHeader("Content-Length", fmt.Sprintf("%d", len(val))))
+		opts = append(opts, apiExpectBody(val), apiExpectContentLength(len(val)))
 	}
 	errs := []error{}
 	if err := a.BlobGetReq(registry, repo, dig, td, opts...); err != nil {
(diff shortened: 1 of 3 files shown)


tianon:
I had implemented transport compression for distribution in https://github.com/distribution/distribution/commit/86f3f1ffec971bbff32cd04549e49f73ce24dbd0, and this found a bug -- it will report a bogus `Content-Length` when answering a `HEAD` request (which is why that's tested explicitly here).

An extra implementor's note would be to be careful to avoid doing transport compression on already-compressed blobs, but it's technically _valid_ to do so, even if mostly useless (and in some cases, actively detrimental to transfer size), but that's *not* explicitly tested here because that's 100% a registry implementation concern, *not* a specification requirement (from either us *or* any of the relevant HTTP specs).

tianon:
> I had implemented transport compression for distribution in https://github.com/distribution/distribution/commit/86f3f1ffec971bbff32cd04549e49f73ce24dbd0, and this found a bug -- it will report a bogus `Content-Length` when answering a `HEAD` request (which is why that's tested explicitly here).
> 
> An extra implementor's note would be to be careful to avoid doing transport compression on already-compressed blobs, but it's technically _valid_ to do so, even if mostly useless (and in some cases, actively detrimental to transfer size), but that's _not_ explicitly tested here because that's 100% a registry implementation concern, _not_ a specification requirement (from either us _or_ any of the relevant HTTP specs).

A second, better attempt in https://github.com/distribution/distribution/commit/8da91c86d8dfbb37dd4e85263303a6bc6bae66ef

jdolitsky (approved):

dmcgowan (approved):

sudo-bmitch (changes requested):
Not a complete rejection, but I'm going to slow this down a bit while we check for possible compatibility issues between clients and registries.

